PR #21, opened 2017-11-02
Add server api reference

### LINKED ISSUE #17: Plugin back-end API reference
Gonna mull over this for another day or two at least before doing anything. Feedback appreciated.

## Step One: Improve Existing Documentation

The existing documentation could be more thorough, and I'd like to add some examples. Examples are a bit weird, because they're expected to be complete, functional, and runnable on the playground. Of course, we can't run plugins on the playground, but I think a decent way to go is to create an `exampleplugin` package that drives the example. It would provide an `exampleplugin.Main` function that makes it interchangeable with `rpcplugin`:

```go
package main

import (
	"fmt"
	"net/http"

        // When building a real plugin, replace "exampleplugin" with "rpcplugin".
	pluginimpl "github.com/mattermost/mattermost-server/plugin/exampleplugin"
)

type MyPlugin struct{}

func (p *MyPlugin) ServeHTTP(w http.ResponseWriter, _ *http.Request) {
	fmt.Fprintf(w, "Hello, world!")
}

func main() {
	pluginimpl.Main(&MyPlugin{})
}
```

`exampleplugin.Main` would invoke whatever hooks the plugin has implemented with representative input, then display some feedback in stdout as necessary to make it clear what's happening.

This would hopefully allow users to get a more hands-on idea of what the inputs and outputs of the API look like.

## Step Two: Integrate the Documentation into the Site

We can use the `godoc` package to parse all of the comments in the standard way: https://godoc.org/golang.org/x/tools/godoc

We more or less just need to make a template and wrap that in a nice little command.

The thing I'm not sure about here is the best way to work this into the build process. Should we just generate site/content/extend/server/api-reference.md entirely before invoking the `hugo` command to generate our HTML? That wouldn't be too bad, but I feel like there's probably a more elegant way to integrate with hugo. (Update: Generating [data templates](https://gohugo.io/templates/data-templates/) instead of markdown would probably be one step better.)

ccbrown:
Make target `go get`s the plugin package and runs a script to generate a JSON representation of the documentation. That JSON is put into the site/data directory, and is used by a layout to generate the HTML for the API reference:

http://mattermost-developer-documentation.s3-website-us-east-1.amazonaws.com/branches/add-server-api-reference/extend/server/reference/

If you just use `hugo server` without generating the docs, that page will just have some placeholder text telling you what to do.

It's just a start, but it's already more usable than GoDoc for this since GoDoc completely neglects documentation for interface methods. In addition to styling and more detailed docs, I'll definitely want an "introduced in version 4.X" tag on each method.

Also updated the Travis settings to build master daily, even if there are no new commits, so these docs will never be more than a day old.

Closes #17

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # build artifacts
 /dist
+/site/data/PluginGoDocs.json
 
 # os artifacts
 *.swp
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,7 @@
 dist: trusty
 sudo: false
 language: go
-go:
-  - 1.x
+go: 1.x
 before_install:
   - mkdir bin
   - export PATH=$PATH:$PWD/bin
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 dist: trusty
 sudo: false
+language: go
+go:
+  - 1.x
 before_install:
   - mkdir bin
   - export PATH=$PATH:$PWD/bin
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,10 @@
-.PHONY: dist
+.PHONY: dist plugin-godocs
 
-dist:
+dist: plugin-godocs
 	rm -rf ./dist
 	cd site && hugo --destination ../dist/html
+
+plugin-godocs:
+	go get -u -v github.com/mattermost/mattermost-server/plugin
+	mkdir -p site/data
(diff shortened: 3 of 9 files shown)


ccbrown:
Added examples, and a shortcode that allows the embedding of example code. So this way, the code used by the hello world tutorial is guaranteed to compile (because it gets compiled every time we run our server tests).

PR #40, opened 2017-11-17
Make use of drafts

jwilander:
* Mark everything but plugin and internal as drafts
* Update navbar to use drafts as well and hide contribute, extend, blog under drafts
* Update homepage to show "Coming Soon" for contribute and integrate sections when not in draft mode
* Disable mailing list until marketing is consulted (might want to switch to mailchimp)

### DIFF
--- a/site/config.toml
+++ b/site/config.toml
@@ -15,6 +15,9 @@ pygmentsStyle = "manni"
   description = ""
   email = ""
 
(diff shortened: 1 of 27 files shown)


ccbrown:
That makes the site start to look pretty presentable as a beta to me.

PR #61, opened 2017-12-06
Update contribution guidelines

jwilander:
Also removed empty page + minor update to paragraph spacing.

### DIFF
--- a/site/content/contribute/_index.md
+++ b/site/content/contribute/_index.md
@@ -1,9 +1,0 @@
-+++
-draft = true
-title = "Contribute"
-date = "2017-08-19T12:01:23-04:00"
-+++
-
-# heading!
-
-text!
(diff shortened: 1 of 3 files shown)


jwilander:
Yeah I was thinking about that as well, but I do think there is a good chunk that would be shared between all of them

We can probably re-org the whole thing a bit to make it work with having some info here and specific info in the other sections

jasonblais:
> We can probably re-org the whole thing a bit to make it work with having some info here and specific info in the other sections

Should I create a separate ticket for it?

jwilander:
@jasonblais yes please, that way this PR won't go stale and we can merge it

jasonblais:
Thanks, I meant to come back to this and review.

Made a couple of minor tweaks https://github.com/mattermost/mattermost-developer-documentation/pull/61/commits/00d161795d3bdbb924463bfc757717783addedc4

PR #73, opened 2018-01-09
Add plugin security page

ccbrown:


jasonblais:
@ccbrown

1) Should we set the expectation that you should always run plugins in a sandbox. E.g. instead of

```
To mitigate the server-side security concerns, Mattermost can run plugin executables in a sandbox on supported platforms. Currently, Linux is the only operating system that Mattermost supports sandboxing for.
```

say

```
Run plugin executables in a sandbox on supported platforms. Currently, Linux is the only operating system that Mattermost supports sandboxing for.
```

2) Ticket to add a note in the System Console when not running plugins in a sandboxed environment. https://mattermost.atlassian.net/browse/ABC-67

It needs design, but created it for us to track.

3) I realize we want to make it clear plugins are powerful and potentially dangerous. But if I was an admin and read this page, I would probably be too scared to enable anything without fully understanding it (which, quite frankly, most won't with the exception of a few admins).

ccbrown:
1. Yeah, probably.

3. This is definitely a concern, but on the other hand, we shouldn't lull people into a false sense of security. :-/

ccbrown:
@jasonblais I changed the wording a bit to encourage readers to use sandboxing. How does this look now?

site/content/extend/plugins/security.md
```diff
@@ -0,0 +1,43 @@
...
+
+# Security
+
+Plugins are powerful. You should only install plugins that you've thoroughly reviewed as they have the potential to compromise the security of your installation.
+
+## Server Considerations
+
+Plugins have the ability to execute arbitrary code on your server. If they aren't properly isolated, they can do just about *anything*. For example, they could read your config file to get your database password, connect to your database, then exfiltrate sensitive user information.
+
+### Sandboxing
```
jasonblais: Should we mention this if for MM server v4.7 and onward?

ccbrown: Probably.

site/content/extend/plugins/security.md
```diff
@@ -0,0 +1,43 @@
...
+
+#### How to Enable Sandboxing
+
+If your platform supports it, it'll be enabled by default. Check your logs for warning messages. If you don't see any, you're all set. If sandboxing is not enabled, you'll see log messages with more details.
+
+If you deploy Mattermost using Docker, sandboxing will be blocked by Docker's default seccomp profile. The easiest way to resolve this is to simply disable the profile with `--security-opt seccomp=unconfined`. Mattermost will still use seccomp for the plugins themselves, so it would be reasonable to just do this and call it a day. But if you really need to run Mattermost with least privilege, you can copy [Docker's default seccomp profile](https://github.com/moby/moby/blob/master/profiles/seccomp/default.json) and add the "clone", "mount", "umount2", and "pivot_root" system calls to the whitelist (We may require others in the future, so you'll need to be vigilant about testing your profile with new releases.).
+
+## Web App Considerations
+
+Plugins have the ability to execute arbitrary code in client browsers. They have the ability to perform nearly any action on behalf of anyone using the web app.
```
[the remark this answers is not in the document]
ccbrown: There's really not anything admins can do right now to mitigate this.

jasonblais:
@ccbrown Added an extra paragraph to the intro section, noting that sandboxing is available in v4.7, and making it clear that only System Admins can upload custom plugins when the one config setting is off.

ccbrown:
@jasonblais 

> Only System Administrators can upload custom plugins and only if `PluginSettings` > `PluginUploads` in config.json is set to `true`.

Unless we make it so administrators can't just enable plugin uploads from the system console (which I do think we should do), I think this is misleading. Until then, as far as the security model is concerned, it's more accurate to just say "Only System Administrators can upload custom plugins.".

jasonblais:
@ccbrown How about:

Only System Administrators can upload custom plugins. They can be disabled in System Console > Plugins > Configuration page.

In Mattermost v4.7 and later, sandboxing adds an additional layer of security for server plugins.

ccbrown:
@jasonblais Yeah I think that’s better.

jasonblais:
@ccbrown Updated

jasonblais:
@ccbrown Updated, good to merge from my end.

PR #77, opened 2018-01-23
Replace +++ in headers with ---

### LINKED ISSUE #70: Switch all file headers to use YAML
We have a mix of YAML (`---`) and TOML (`+++`) in page headers. Let's switch them all to use YAML for consistency's sake.

Hanzei:
Fixes #70

### DIFF
--- a/site/content/contribute/devtalks/_index.md
+++ b/site/content/contribute/devtalks/_index.md
@@ -1,8 +1,8 @@
-+++
-title = "Mattermost Dev Talks"
-date = "2018-01-05T09:51:35-05:00"
-section = "contribute"
-+++
+---
+title: "Mattermost Dev Talks"
+date: "2018-01-05T09:51:35-05:00"
+section: "contribute"
+---
 
 # Mattermost Dev Talks
 Approximately once a month, a member of the Mattermost team releases a developer talk that discusses the architecture, design principles, and tools that the Mattermost project uses.
--- a/site/content/contribute/getting-started/_index.md
+++ b/site/content/contribute/getting-started/_index.md
@@ -1,9 +1,9 @@
-+++
-draft = true
-title = "Getting Started"
-date = "2017-08-19T12:01:23-04:00"
-section = "contribute"
-+++
+---
+draft: true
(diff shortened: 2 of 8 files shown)


Hanzei:
The build failed, but I'm quite sure its not related to my changes.
```
go run scripts/devtalks.go > site/data/DevTalks.json
2018/01/23 13:34:21 Environment variable YOUTUBE_API_KEY must be set in order to fetch dev talks
exit status 1
make: *** [devtalks-data] Error 1
```

jwilander:
@MusikPolice any idea why the build is failing on that?

site/content/contribute/devtalks/_index.md
```diff
@@ -1,12 +1,12 @@
-+++
+---
 title = "Mattermost Dev Talks"
```
jwilander: I think you also need to change the `=`s to `:`s, so that's it YAML formatted

Hanzei: Your right. I will fix this.

Hanzei:
Since we use YAML in all files, should we use it also for the config file which is using `toml` atm?

MusikPolice:
@jwilander I'm not sure. The environment variable is set in the build configuration.

MusikPolice:
Hmm the build issue might be because the environment variable is secret, and @Hanzei is pushing the PR from a fork of the repository. [This documentation](https://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-Repository-Settings) seems to imply that secret environment variables won't be applied to builds of repository forks: 
> By default, the value of these new environment variables is hidden from the export line in the logs. This corresponds to the behavior of encrypted variables in your .travis.yml. The variables are stored encrypted in our systems, and get decrypted when the build script is generated.
> Similarly, we do not provide these values to untrusted builds, triggered by pull requests from another repository.
Suspect that even if that script passed, we would fail to deploy to AWS

jwilander:
I'll test it locally quickly and then merge

jwilander:
Works great 👍

PR #80, opened 2018-01-25
Adding bizible script to header

asaadmahmood:
Hey guys, adding the bizible script to the head portion of the site on the request of Benoite.

### DIFF
--- a/site/layouts/partials/head.html
+++ b/site/layouts/partials/head.html
@@ -24,3 +24,4 @@
     <script src="{{ .Site.BaseURL }}/js/tabs.js"></script>
     <script src="{{ .Site.BaseURL }}/js/main.js"></script>
     <script async defer src="https://buttons.github.io/buttons.js"></script>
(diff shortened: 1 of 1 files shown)


site/layouts/partials/head.html
```diff
@@ -24,3 +24,4 @@
     <script src="{{ .Site.BaseURL }}/js/tabs.js"></script>
     <script src="{{ .Site.BaseURL }}/js/main.js"></script>
     <script async defer src="https://buttons.github.io/buttons.js"></script>
+    <script type="text/javascript" src="//cdn.bizible.com/scripts/bizible.js" async=""></script>
```
ccbrown: The MIME type `text/javascript` is obsolete (https://tools.ietf.org/html/rfc4329#section-7.1). Could just remove that attribute.

I would also be a fan of using the minimized attribute form for async like the line above it.

asaadmahmood:
Hey @ccbrown as far as I am told, this is to track where the users go on our page.
https://www.bizible.com/marketing-attribution-for-marketing-operations-leaders

PR #81, opened 2018-01-25
Add a new sample plugin to dev docs site

jasonblais:


ccbrown:
The documentation looks good to me, but that plugin really isn't an example at all. You can't easily look at the source code, and even if you did, you wouldn't find anything at all relevant to plugin development since it uses a very different mechanism.

Is there somewhere else we can put this? Alternatively, maybe we should just prioritize porting this plugin to the RPC mechanism like the others.

jasonblais:
@ccbrown Hm, yeah. Good point. Maybe I should add it to our docs.mattermost.com page instead? Since it's really meant for admins rather than developers.

ccbrown:
Makes sense to me.

jasonblais:
PR in docs: https://github.com/mattermost/docs/pull/1754/

PR #82, opened 2018-01-26
Added instructions for regenerating pfx files

MusikPolice:


site/content/contribute/desktop/code-signing.md
```diff
@@ -16,6 +16,24 @@ In order to code sign releases on behalf of Mattermost Inc., you'll need a `.pfx
...
 
+#### Regenerating the .pfx File
+Signing certicates occasionally expire. When renewed, the registrar will provide you with  a `.spc` (Software Publishing Certificate) and a `.pem` (Public Key) file. These must be combined with a corresponding `.key` (Private Key) file to create the `.pfx` (Personal Information Exchange) file that is used to sign builds.
+
+First, verify that the `.pem` and `.key` files match:
+```bash
+~$ openssl x509 -noout -modulus -in public.pem | openssl md5
+ba31f9e34b7ef88af514b0a23f1b72b8
+~$ openssl rsa -noout -modulus -in private.key | openssl md5
+ba31f9e34b7ef88af514b0a23f1b72b8
```
MusikPolice: I changed these hashes so that we weren't leaking our real ones, just in case that's a bad idea (I don't think it is, because the public key is public, but best to be safe about it)

MusikPolice:
@ccbrown The uploaded image was already full sized. I increased its size on the page to make it easier to read.

PR #83, opened 2018-02-06
Bus Error Avoidance Documentation

MusikPolice:
Added documentation for avoiding a bus error when codesigning Windows executables on Linux systems


MusikPolice:
Merged with one review, because it's just documentation

PR #91, opened 2018-03-04
Add triangle button in top-right of navbar linking to internal docs

jwilander:
Makes it easier to get to the internal documentation

### DIFF
--- a/site/layouts/partials/nav.html
+++ b/site/layouts/partials/nav.html
@@ -14,6 +14,7 @@
             {{ end }}
         {{ end }}
     </ul>
+    <div class="nav-triangle-box"><a class="nav-triangle-link nav-triangle-link-home" href="/internal"><i class="nav-i nav-i-home fa fa-info"></i></a></div>
 </header>
 
 {{ partial "notification.html" .}}
--- a/site/layouts/partials/nav2.html
+++ b/site/layouts/partials/nav2.html
@@ -17,6 +17,7 @@
             {{ end }}
         {{ end }}
     </ul>
+    <div class="nav-triangle-box"><a class="nav-triangle-link" href="/internal"><i class="nav-i fa fa-info"></i></a></div>
 </header>
 
 {{ partial "notification.html" .}}
--- a/site/static/css/styles.css
+++ b/site/static/css/styles.css
@@ -2686,3 +2686,36 @@ table > tbody > tr:nth-of-type(odd) {
   .homepage-boxes {
     margin-top: -13em; } }
 
+.nav-triangle-box {
+  position: absolute;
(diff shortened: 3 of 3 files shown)


jasonblais:
It's not really obvious. My first intuition was that the "i" triangle button gives info about what the dev site is, or some other important information relevant to me.

Maybe check with AM on his thoughts?

ccbrown:
Yeah, I agree with @jasonblais. This definitely looks like it should open an "about this site" page or something similar. Maybe a more subtle footer link would be a good alternative.

jwilander:
The point was kind of to be non-obvious but yeah I guess it can be pretty misleading, so let's find a different solution

jasonblais:
@jwilander Maybe we add a footer for the site, which would also contain a link to the internal docs?

jwilander:
A footer would be ok, but then I need to scroll all the way to the bottom of the page to click on it so it's not very fast IMO

jwilander:
The main sections don't actually show up in the LHS. I mean we can just add
the Internal link to the header, I don't think it's a huge deal. The only
downside is that more people will see it when it won't be much use them and
probably won't be as well written either. Hence me breaking our design
principle and making it non-obvious.

If we're ok with non-obvious we could do something like I did here except
it doesn't show up at all until you hover over it

jasonblais:
How about a link to internal docs via the Dev channel header on pre-release?

Still, just brainstorming options, 0/5. But doing something non-obvious is probably not ideal.

jwilander:
How is not having a link on the site at all any less obvious than having a hidden one? I think our "fast, obvious, forgiving" design principle is great but I don't think we need to hold to it as the letter of god, so to speak. I think we can definitely make conscious decisions to break from it special circumstances when it makes sense.

The goal for me here is to make it easy and fast for a staff developer to get to the internal documentation from anywhere on the dev docs site, while at the same time hiding it from the general user of the site (not because of anything confidential, just because it's noise they don't need to worry about). So to me it makes sense to make it non-obvious how to get there upfront then, with a tiny bit of training for staff, make it easy to access when you do know about it.

Sure, that directly contradicts with part of the design principle but in this case I think it makes sense to break from it a little to properly meet the goal.

We can bring it up with Mike if you like, but I'm definitely 4/5 on this

jasonblais:
I was referring to external developer/customer earlier when taking about obviousness. Not staff.

If we have the original design that only shows up on hover with words like `Core Team`, it's less intrusive + more obvious what that link takes you when clicked (some docs relevant to the Core Team).

As above, I'm 0/5. If you're 4/5 on something, we can just go ahead and see how it feels when it's live.

PR #98, opened 2018-04-11
MM-10126/MM-10165 Document mainMenuActions and exported libraries

jwilander:
https://mattermost.atlassian.net/browse/MM-10126 ~~plus some extra documentation of exported libraries~~ and https://mattermost.atlassian.net/browse/MM-10165

### DIFF
--- a/site/content/extend/plugins/webapp/reference.md
+++ b/site/content/extend/plugins/webapp/reference.md
@@ -132,7 +132,7 @@ Contains the following post utility functions:
 ##### `formatText(text, options)`
 Performs formatting of text including Markdown, highlighting mentions and search terms and converting URLs, hashtags, @mentions and ~channels to links by taking a string and returning a string of formatted HTML.
 
-* `text` - String of text to format, e.g. a post's message
+* `text` - String of text to format, e.g. a post's message.
 * `options` - (Optional) An object containing the following formatting options
  * `searchTerm` - If specified, this word is highlighted in the resulting HTML. Defaults to nothing.
  * `mentionHighlight` - Specifies whether or not to highlight mentions of the current user. Defaults to true.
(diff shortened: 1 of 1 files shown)


jasonblais:
Nice! I'll review this soon. 

Looks like this would cover https://mattermost.atlassian.net/browse/MM-10165 as well that I created earlier today.

jwilander:
Oh yeah, guess it would

site/content/extend/plugins/webapp/reference.md
```diff
@@ -96,3 +111,80 @@ The theme object has the following properties:
 | mentionHighlightLink | Color of text for mention links in posts |
 | codeTheme | Code block theme, either 'github', 'monokai', 'solarized-dark' or 'solarized-light' |
 
+## Exported Libraries and Functions
+
+The web app exports a number of libraries and functions on the [window](https://developer.mozilla.org/en-US/docs/Web/API/Window) object for plugins to use. We recommend importing as many libraries from the window as possible. Below is a list of the exposed libraries and functions:
+
+
+| Library | Description |
+| -------- | ----------- |
```
jasonblais: I'm wondering if it's useful to describe when you'd import a library, and when not? E.g. use `react` library if you do X and Y?

site/content/extend/plugins/webapp/reference.md
```diff
@@ -96,3 +111,80 @@ The theme object has the following properties:
 | mentionHighlightLink | Color of text for mention links in posts |
 | codeTheme | Code block theme, either 'github', 'monokai', 'solarized-dark' or 'solarized-light' |
 
+## Exported Libraries and Functions
```
jasonblais: Do you mean `Exposed` libraries?

jwilander: Nah I wanted `Exported` but I do tend to use it interchangeably with exposed

site/content/extend/plugins/webapp/reference.md
jwilander: Maybe, might be a bit overkill though. If you need a library as a developer you should know why and if you don't need then you can ignore these. Linking out is more then enough IMO

site/content/extend/plugins/webapp/reference.md
```diff
@@ -96,3 +111,80 @@ The theme object has the following properties:
...
+ * `searchTerm` - If specified, this word is highlighted in the resulting HTML. Defaults to nothing.
+ * `mentionHighlight` - Specifies whether or not to highlight mentions of the current user. Defaults to true.
+ * `mentionKeys` - A list of mention keys for the current user to highlight.
+ * `singleline` - Specifies whether or not to remove newlines. Defaults to false.
+ * `emoticons` - Enables emoticon parsing with a data-emoticon attribute. Defaults to true.
+ * `markdown` - Enables markdown parsing. Defaults to true.
+ * `siteURL` - The origin of this Mattermost instance. If provided, links to channels and posts will be replaced with internal links that can be handled by a special click handler.
+ * `atMentions` - Whether or not to render "@" mentions into spans with a data-mention attribute. Defaults to false.
+ * `channelNamesMap` - An object mapping channel display names to channels. If provided, ~channel mentions will be replaced with links to the relevant channel.
+ * `team` - The current team.
```
lieut-data: Would this be a `teamId`?

site/content/extend/plugins/webapp/reference.md
```diff
@@ -96,3 +111,80 @@ The theme object has the following properties:
...
+ * `proxyImages` - If specified, images are proxied. Defaults to false.
+
+##### `messageHtmlToComponent(html, isRHS, options)`
+Converts HTML to React components.
+
+* `html` - String of HTML to convert to React components
+* `isRHS` - Boolean indicating if the resulting componets are to be displayed in the right-hand sidebar
```
lieut-data: typo: `components`

Also, it's not super clear to me why I'd use this flag. Does passing `isRhs: true` make the components automatically appear in the RHS? Does HTML in the RHS require special treatment?

lieut-data:
Perfect 👍

PR #103, opened 2018-04-23
Stop hiding integrate section, add integrate example section, other minor changes

jwilander:
Finish off the last pieces of the integrate section 🎉 

@jasonblais I seeded the examples page with some integrations. Not sure if we want to add images to liven it up a little or other popular integrations there, thoughts?

Everyone: please take a look through whole integrate section (not just what changed here) and feel free to give feedback if you think something needs to change before the page goes public


jasonblais:
Why were these integrations chosen? Any drivers that would be useful to reference? There are also others we can probably add - queued for our 1-1 tomorrow.

Are we also going to do updates to docs.mattermost.com at the same time before/while publishing?

crspeller:
Couple things:
- The main page where we link to Developer Toolkit we should have a list of usefull links to channels.
- We should have a cross link to the plugins section.

jwilander:
@lieut-data thanks for the feedback!

1. Generally because I generally think of the Developers channel more for people contributing directly to Mattermost and the Developer Toolkit channel is more focused on integrations. Though maybe it makes sense to link to both
2. Whoops, I meant to redirect that to the admin site until we add specifics docs for that here. Will update.
3. I think your way makes sense. Would you be willing to push a commit to this PR to fix that?
4. I'm going to be honest I can't see the difference :P but I'm definitely OK with changing it

jwilander:
@crspeller thanks as well!

1. Agreed, will update.
2. We do link to the plugins in the `How Should I Integrate` section, where were you thinking of adding the link?

crspeller:
2. Ya I missed it at the bottom there.

lieut-data:
@jwilander 
1. 👍 
2. 👍 
3. Will do!
4. This is primarily what drove the suggestion:

image

If you're amenable, I'll include a patch for this as well.

jwilander:
re 4): Ah yes, I see what you mean. If you want to update that'd be great 👍

jasonblais:
@jwilander This is really good. Excited to go live! Accompanying doc PR to replace some integrations content with links to these new developer docs sitehttps://github.com/mattermost/docs/pull/1978

Two notes:

1) Do we want to cross-link to developers.mattermost.com somewhere in api.mattermost.com?
2) If I first go to ../extend/plugins/, then hit the "Integrate" section in the header, I'm redirected to ../extend/plugins/integrate/getting-started, instead of ../integrate/getting-started. Similarly if I go to the Integrate section first and then to Extend in the channel header

Few notes on webhook and slash command pages that I missed on my earlier review:
3) For incoming webhook parameter table, do we need to mention we expect the channel handle for the `channel` parameter, not the display name? Something like

```
Use the channel URL, not the channel display name, when specifying the channel parameter. For instance, use town-square, not Town Square when posting messages to the Town Square channel.
```

Some people have hit issues with this in the past.

4) For incoming/outgoing webhook and slash command pages, should we mention the different types of mentions supported. E.g.

```
You can trigger mention notifications with your incoming webhook message. To trigger a mention, include *@username* or *<userid>* in the `text` parameter of the JSON payload.

Channels can be mentioned by including *@channel* or *<!channel>*. For example:

payload={"text": "<!channel> this is a notification."}
```

5) Are any of the tips & best practices, FAQ or troubleshooting items from https://docs.mattermost.com/developer/webhooks-incoming.html useful for incoming webhooks?
6) Same as 5 except for outgoing webhooks https://docs.mattermost.com/developer/webhooks-outgoing.html
7) `response_url` and `goto_location` parameters are missing from slash commands doc. Here are the excerpts from docs.mattermost.com

```
Redirect to an external webpage
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

You can use the ``goto_location`` parameter to redirect the user of your command to an external webpage.

If the parameter is present, the user is redirected to the specified URL in a new browser tab. For instance, to redirect the user to `https://about.mattermost.com <https://about.mattermost.com>`_, use the following response.

.. code-block:: text

  {"response_type": "ephemeral", "goto_location": "https://about.mattermost.com", "text": "Hello, this is some text\nThis is more text. :tada:"}

The parameter supports any custom protocol including ``http://``, ``https://``, ``ftp://``, ``ssh://`` and ``mailto://``.

Delayed and multiple responses
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

You can use the ``response_url`` parameter to supply multiple responses or a delayed response to a slash command. Response URLs can be used to send five additional messages within a 30-minute time period from the original command invocation. 

Delayed responses are useful when the action takes more than three seconds to perform. For instance:
- Retrieval of data from external third-party services, where the response time may take longer than three seconds.
- Report generation, batch processing or other long-running processes that take longer than three seconds to respond.

Any requests that are made to the response URL should either be a plain text or JSON-encoded body. The JSON-encoded message supports both Markdown formatting and message attachments.
```

8) docs.mattermost.com say [response_type is required](https://docs.mattermost.com/developer/slash-commands.html#command-with-a-trigger-of-trigger-word-returned-an-empty-response) for slash commands, but devs site says it's not. Want to confirm what we've got here is accurate.

site/content/integrate/examples/_index.md
```diff
@@ -0,0 +1,29 @@
+---
+title: "Examples"
+date: "2017-01-19T12:01:23-04:00"
+section: "integrate"
+---
+
+# Examples
```
jasonblais: I'll try to look at other integrations we can highlight. Another one that comes to mind is https://github.com/mattermost/mattermost-bot-sample-golang

site/content/integrate/examples/_index.md
```diff
@@ -0,0 +1,29 @@
...
+
+Backend code for powering a Mattermost slash command that creates an interactive poll. Contains Firebase Cloud Functions that connect Mattermost Interactive Buttons with the Firebase Realtime Database.
+
+### [matterbridge](https://github.com/42wim/matterbridge)
+
+Simple bridge between Mattermost, IRC, XMPP, Gitter, Slack, Discord, Telegram, Rocket.Chat, Hipchat(via xmpp), Matrix, Steam and ssh-chat.
+
+---
+
+Find many more integrations <a target="_blank" href="https://about.mattermost.com/community-applications/">here</a>!
```
jasonblais: Maybe have this near the top rather than at the bottom

site/content/integrate/examples/_index.md
jasonblais: Actually, these integrations are good examples. Only other one that comes to mind is the Golang bot https://github.com/mattermost/mattermost-bot-sample-golang

I'm wondering if we should make the title more descriptive, e.g. `Hubot adapter for Mattermost` and reference the project name and link in the description. Something like `MattermostOnFire` isn't obvious on first ready for instance. Thoughts?

jwilander: Yeah, agreed titles could be better. Feel free to update them if you want (if I don't get to it first)

jwilander:
1. Yes, probably somewhere in the introduction of api.mattermost.com
2. I'll fix that
3. Probably a good idea, I'll add a note about it
4. Yes we probably should add it but in the text you pasted I'm not sure what the userid part of this means, I don't think we have mentions by user ID do we? `*@username* or *<userid>*`
5. Yes, most of those would be useful. I'll add them with some slight tweaks to orient it to devs
6. Same as 5
7. Will add
8. It's not required, but it looks the default is `ephemeral` not `in_channel` like the doc says, eg. https://github.com/mattermost/mattermost-server/blob/master/app/command.go#L55. I'll update that

jwilander:
@jasonblais updates made.

Regarding 7, I couldn't find any reference to a `response_url` in the code so unless I'm looking in the wrong place I don't think we support that. Not sure how we added documentation for a non-existent feature

jasonblais:
Looks good! Just two notes:

3 - I don't see this note
4 - For mentions we support `@username` or `<userid>`, and `@channel` or `<!channel>` (similarly for all/here). The `<...>` pieces are for Slack-compatibility, so might be worth mentioning.

Note that by `@username`, I mean something like `@[username]`. `<userid>` actually contains the less/greater than signs inside the message.

jasonblais:
Whoops, didn't mean to close this.

jasonblais:
3 - Looks like I missed it.
4 - I added them to the Slack-compatibility sections

lieut-data:
This is supremely awesome 🎉

PR #104, opened 2018-05-11
Add notes how to install plugins when not uploading via UI

jasonblais:


site/content/extend/plugins/server/hello-world.md
```diff
@@ -48,8 +48,9 @@ You should now have a file named "plugin.tar.gz" in your workspace. Congratulati
 
 ## Installing the Plugin
 
-Open up Mattermost as a system administrator, navigate to **Plugins > Management** and upload the `plugin.tar.gz` you generated above.
+Install the plugin in one of the following ways:
 
-Click "Activate" under the plugin when it's done uploading.
+ - Log in to Mattermost as a System Admin, navigate to **Plugins > Management** and upload the `plugin.tar.gz` you generated above. Then click "Activate" under the plugin after it has uploaded.
+ - Add the plugin to the directory set by **PluginSettings > Directory** in your ``config.json`` file and restart the server
```
jwilander: That setting could be missing if it's from someone with an older config. It defaults to `./plugins` if it's not found

Also we should clarify that the plugin needs to be extracted into a folder named the plugin id of the plugin. It would look like this for the hello world example:

```
mattermost/
    plugins/
        com.mattermost.server-hello-world/
            plugin.yaml
            plugin.exe
```

jasonblais: How do the updates look like @jwilander ?

jasonblais: (I will make them for the webapp one too if it looks okay)

jwilander: Looks good, only difference is that the webapp structure would be:

```
mattermost/
    plugins/
        helloworld/
            webapp/
                helloworld_bundle.js
```

And the webapp plugin id is `helloworld`

jasonblais:
Updated.

PR #105, opened 2018-05-11
Update documentation re: GET and slash commands.

lieut-data:
This complements https://github.com/mattermost/mattermost-server/pull/8779.

@jwilander, do we need to distinguish versions of the Mattermost software in this documentation?

### DIFF
--- a/site/content/integrate/slash-commands/_index.md
+++ b/site/content/integrate/slash-commands/_index.md
@@ -131,6 +131,10 @@ Best practice suggests using `GET` only if a request is considered idempotent. T
 
 Ultimately, however, the choice is yours. If in doubt, configure your slash command to use a `POST` request.
 
+<div class="alert alert-warning" role="alert">
+Note that slash commands configured to use a <code>GET</code> request were broken prior to Mattermost release 5.0.0. The payload was incorrectly encoded in the body of the <code>GET</code> request instead of in the query string. While it was still technically possible to extract the payload, this was non-standard and broke some development stacks.
+</div>
+
 #### Why does my slash command always fail with `returned an empty response`?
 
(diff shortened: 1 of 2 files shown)


site/content/integrate/slash-commands/_index.md
```diff
@@ -127,9 +127,9 @@ During development, you may override this behaviour by setting `ServiceSettings.
 
 #### Should I configure my slash command to use `POST` or `GET`?
 
-While it is possible to configure a slash command to emit a GET request, the corresponding payload sent by Mattermost is emitted as part of the body instead of being sent in the query string. This is non-standard, may cause problems with some software stacks and is not recommended at this time. 
+Best practices suggest using `GET` only if a request is considered idempotent. This means that the request can be repeated safely and can be expected to return the same response for a given input. Some servers hosting your slash command may also impose a limit to the amount of data passed through the query string of a `GET` request.
 
-Additionally, this feature may be deprecated from Mattermost in the future.
+Ultimately, however, the choice is yours. If in doubt, configure your slash command to receive a `POST` request.
```
jespino: Here the slash command emits requests, not receives. But the problem is this is a matter of perspective, what is the slash command, the server configured slash command, or the actual server that reply to the slash command? I would use, the same terminology in the other paragraph and in the title (use instead of receive), something like "If in doubt, configure your slash command to use a `POST` request"

lieut-data: Excellent points! Fixed :)

jwilander:
Yeah, I'd just add a small note about the minimum version. I'm hoping to be able to avoid having a comprehensive versioning of the docs since we release often. So unless anyone feels strongly the other way I figure just using small notes where possible is good enough

lieut-data:
I've updated with a warning. Thoughts?

image

PR #109, opened 2018-05-28
Add vpn setup

cpanato:
Adding post in how to configure the VPN

ping: @jasonblais 

I dont know what is the process for this type of post


jwilander:
Looks good, FYI there's an issue the build where it fails on forks

PR #113, opened 2018-06-05
Add vpn setup

cpanato:


site/content/internal/infrastructure/vpn.md
```diff
@@ -0,0 +1,33 @@
+---
+title: "VPN"
+date: 2018-05-28T16:08:19+02:00
+draft: true
```
jwilander: Actually I think you want to remove `draft: true` so it'll show up on the live site

cpanato: done

PR #125, opened 2018-06-22
document how to use the openvpn cli

lieut-data:
I realize this doesn't yet work reliably outside of OSX, but documenting it here for posterity anyway.

image


cpanato:
I use the UI client :)

PR #153, opened 2018-08-27
Add short plugin overview and steps to add an API to contribution docs

jwilander:
Wanted to get us started on contribution docs for plugins.

### DIFF
--- a/site/content/contribute/server/plugins.md
+++ b/site/content/contribute/server/plugins.md
@@ -17,7 +17,7 @@ Looking to build a plugin? [Then you want the plugin author documenation.](/exte
 
 ## Overview
 
-Plugins are generally made of at least three parts: a manifest, a server binary and a JavaScript bundle.
+Plugins are generally made of at least two parts: a manifest and a server binary and/or a JavaScript bundle.
 
 The manifest tells Mattermost what the plugin is and provides a set of metadata used by the server to install and run the plugin. For more information you can see the [manifest reference](/extend/plugins/manifest-reference). Manifests may be defined in JSON or YAML.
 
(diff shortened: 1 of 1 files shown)


site/content/contribute/server/plugins.md
```diff
@@ -0,0 +1,42 @@
...
+
+Check out the [`plugin` package](https://github.com/mattermost/mattermost-server/tree/master/plugin) and the [plugin_* files in the `app` package](https://github.com/mattermost/mattermost-server/tree/master/app) for the code, and [mattermost-plugin-demo](https://github.com/mattermost/mattermost-plugin-demo) for an example plugin.
+
+## Adding an API 
+
+To add a plugin API you need to add the signature of your new method to the [API interface](https://github.com/mattermost/mattermost-server/blob/master/plugin/api.go). You then need to implement the API in the [plugin_api.go](https://github.com/mattermost/mattermost-server/blob/master/app/plugin_api.go) of the `app` package. Finally, you would need to run `make pluginapi` to generate the RPC glue code needed for your new API.
```
jasonblais: Are any (unit) tests added?

jwilander: Testing of the wireframe is covered by the auto-generated code. For more complex API implementations we'll want separate tests but that'll be the special case so we'll ask for it case-by-case generally

site/content/contribute/server/plugins.md
```diff
@@ -0,0 +1,42 @@
...
+
+In the future there will be another component of plugins for the mobile apps, likely a React Native bundle.
```
crspeller: Maybe add "for mobile". Some people interested might not know that our mobile app is React Native.

jwilander: It does say "for the mobile apps" right before, is that not good enough?

enahum: It is clear to me.

One thing that I’m thinking is that we should add somewhere that if you are making a plugin to change the format For a custom post type the mobile apps won’t render them as expected but instead will render the raw message formatted with our standard formatter?

PR #154, opened 2018-08-30
Update developer-workflow.md

JustinReynolds-MM:


jasonblais:
Carlos and I reviewed our existing developer's guide and we found these two would be helpful to include in the dev docs.

Justin helped add it to the developer workflow, but other suggestions are welcome.

PR #159, opened 2018-09-16
[MM-11293] Documentation for Megacheck Make Target

DSchalla:
Documentation for the megacheck make target including some guidance.

CC @jwilander and @jasonblais


DSchalla:
Thanks  @jasonblais

PR #164, opened 2018-09-23
Support ellipsis args

kaakaa:
Parsing godoc scripts doesn't handle variable length arguments (Ellipsis type).
So `LogXXX` functions should have variable length arguments of `internface{}`, but it is not reflected in the document.
https://developers.mattermost.com/extend/plugins/server/reference/#API.LogDebug

This PR fix it.


jasonblais:
@lieut-data Is this one ready to merge?

PR #172, opened 2018-10-06
Dev Docs Updates

jasonblais:
This PR contains proposed updates for
1. Mac server developer setup guide
2. Webapp developer setup guide

Once the propsed changes for 1) are approved, I can make similar changes for the guides on other platforms.

List of changes:
1. Mac server developer setup guide
 - Use consistent language throughout, e.g. `following` vs `using` when referencing instructions on other pages
 - Updated steps for editing the `/etc/hosts` file [if this is overkill for devs, let me know]
 - Combined steps 7 (running and stopping Mattermost server) and 8 (testing install) to a single step. Currently if you run step 7 first, you end up stopping the server, and thus step 8 will automatically fail
 - Changed https://localhost:8065 to http. https gives a `This site can’t provide a secure connection` error. 
2. Webapp developer setup guide
 - Updated pre-requisites; e.g. moved setting up server environment and installing libpng to pre-requisites
 - Modified setup steps to have consistent language with the server setup guide


site/content/contribute/webapp/developer-setup.md
```diff
@@ -5,24 +5,35 @@ weight: 2
...
+#### Prerequisites
 
-## Setup
+1. Set up your [development environment for the Mattermost server](/contribute/server/developer-setup). 
 
-Follow these steps to set up your web app developer environment:
+ - Running these steps will also install Git on your machine, and Homebrew for Mac.
 
-1. To develop for the web app you're also going to need to [set up the developer environment for the server](/contribute/server/developer-setup). So if you haven't done that yet, start there and then come back here.
+2. Install Node.js and npm following the instructions at https://www.npmjs.com/get-npm.
```
zeroimpl: I suggest specifying the min required version. The nodejs website actually recommends downloading 8.12.0, but that fails the unit tests. Perhaps:
```2. Install Node.js and npm following the instructions at https://www.npmjs.com/get-npm, using node version 10.x or later.```

File: site/content/contribute/webapp/developer-setup.md [hunk 2]
```diff
@@ -5,24 +5,35 @@ weight: 2
...
-3. If you're on Mac, you'll need to install libpng as well: `brew install libpng`
+    ```sh
+    brew install libpng
+    ```
 
-4. Go to https://github.com/mattermost/mattermost-webapp and fork the repository.
+#### Setup
 
-5. Clone your fork `git clone https://github.com/<yourgithubusername>/mattermost-webapp` beside your `mattermost-server`.
+1. Go to https://github.com/mattermost/mattermost-webapp and create a fork.
```
lieut-data: The location of the fork is important: various scripts currently assume it to be a sibling to the `mattermost-server` directory. Also, there's been some recent confusion about a missing symlink. Suggest:

1. Go to https://github.com/mattermost/mattermost-webapp and create a fork.
2. Clone the fork next to your mattermost-server directory, replacing `{yourgithubusername}` with your GitHub username:
```sh
git clone https://github.com/{yourgithubusername}/mattermost-webapp.git $(go env GOPATH)/go/src/mattermost/mattermost-webapp
```
3. Link your server with the webapp:
```
ln -s $(go env GOPATH)/src/github.com/mattermost/mattermost-webapp/dist $(go env GOPATH)/src/github.com/mattermost/mattermost-server/client
```
4. Test your environment:
```sh
cd $(go env GOPATH)/go/src/mattermost/mattermost-webapp
make test
```

jasonblais:
Thank you @lieut-data! Appreciate your feedback.

Out of curiosity, do we have a target audience in mind for developer docs? E.g. a developer with basic experience working with Docker and the Go workspace?

jasonblais:
Thanks all for the feedback. Updated

MartinDelille:
I proposed a better (in my opinion) configuration here: https://github.com/mattermost/mattermost-server/blob/915a2b039d8732a3db01623ff5ebf0a5f35fdf54/doc/mac.md

- I use `go get` to clone the project which automate the directory creation and the cloning
- I use `git remote` which allow to pull easily further improvement from the upstream repo

Are you interested into this?

site/content/contribute/webapp/developer-setup.md
MartinDelille: why not using `go get`?

lieut-data: I could be mistaken, but I don't think `go get` works with non-Go repositories? We'll eventually adopt Go modules, at which point `go get` isn't the right path forward, since modules means pivoting away from `$GOPATH`.

lieut-data:
@MartinDelille, thanks for the idea! I do agree that `go get` helps automate the directory hierarchy, and it makes sense to think about this for the server repository. As noted above, however, I don't think this works for non-Go repositories (like mattermost-webapp), and I don't think this fits into a longer-term Go modules plan that would obviate the need for the `$GOPATH` making it easier to co-locate the dependencies as such.

I don't work out of a fork, but if you clone your fork, is it common to set another upstream for our repository to pull down changes?

site/content/contribute/webapp/developer-setup.md
MartinDelille: It works: I tested it. The only problem is that it trigger a warning and at the beggining I thought it failed.

Why mattermost-server isn't a Go repository?

hanzei:
About the second upstream: E.g. hugo does this, see https://github.com/gohugoio/hugo/blob/master/CONTRIBUTING.md#fetching-the-sources-from-github.

I use a second upstream myself and it works like a charm. You don't longer have to sync masters.

MartinDelille:
@lieut-data Yes adding an upstream remote to pull down changes work also. This is not the way I do but I totally understand that you do it another way. You can just ignore my proposition I won't be upset don't worry! 😄

jasonblais:
@lieut-data Have you got around to reviewing the latest changes on this PR?

lieut-data:
Ah, nope. Disappeared from my radar altogether -- re-requesting my review to put it back.

lieut-data:
@jasonblais, I've pushed some proposed simplifications. Can I get your eyes on same? If you like them, I can apply the relative delta to the other operating systems in the server block.

lieut-data:
Sounds good, and will do!

lieut-data:
@jasonblais, I've updated the other sections (breaking them out into discrete files). I've also run through all the Linux variations, but I don't have the ability to check the Windows variants at this time.

jasonblais:
Thank you @lieut-data! Much appreciated 👍 Merging

PR #173, opened 2018-10-07
Fix crash when running plugin-jsdocs.js

enahum:
Don't really know why the failure is pointing at my commit but it seems that somehow `statement.loadingComments` is undefined or null and thus causing a crash when running the script, this PR makes it so it does not crash.


enahum:
Should we merge this?

PR #174, opened 2018-10-15
Update developer-setup.md

btotharye:
Updated typo for http and not https since it doesn't have TLS out of the box.


btotharye:
Closing as duplicate of #172

PR #184, opened 2018-11-06
Add k8s documentation

cpanato:
In draft mode for now to get feedback

### DIFF
--- a/site/content/internal/infrastructure/_index.md
+++ b/site/content/internal/infrastructure/_index.md
@@ -2,6 +2,7 @@
 title: Infrastructure
 date: 2017-11-06T19:30:07-05:00
 section: internal
+weight: 10
 ---
 
 Select a topic to the left.
--- a/site/content/internal/infrastructure/aws.md
+++ b/site/content/internal/infrastructure/aws.md
@@ -2,6 +2,7 @@
 title: AWS
 date: 2017-11-06T19:30:07-05:00
 subsection: internal
+weight: 40
 ---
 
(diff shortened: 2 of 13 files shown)


cpanato:
@jasonblais @jwilander please let me know what I need to add/remove/rewrite...

cpanato:
@jasonblais 
1.  Would we want to include a note on how to uninstall the chart (e.g. if something went wrong and the dev/admin wants to start over)?

in our case, for now, we are not using helm, but I will add a note in how to remove using helm and using this approach

2. Do we have a comparable list of config parameters similar to Team Edition?

Do we need to put here? I think this is more case by case or use the default

3. Is there any functionality in Mattermost that is not supported on Kubernetes (I assume not, but wanted to confirm)?

humm, i guess the answer is no.

4. Can you choose any database?

will add this section. was missing.

5. Any known limitations?

will add as well

jasonblais:
1 + 2: Right, sorry, I was mixing helm and kubernetes when reading the TE kubernetes doc. So I think those are okay as-is.

cpanato:
@jasonblais @jwilander PTAL

jasonblais:
@jwilander Asked 👍

cpanato:
@jwilander thanks 😄 

I was thinking to write a blog post, but I'm not good at writing things :( but this will be a good thing to do

@jasonblais

jwilander:
It'll be good practice and Justin can help you with the wording/grammar :)

PR #187, opened 2018-11-14
Add note about testify in go tests

hanzei:
2/5 on the wording.


jwilander:
Wording looks good to me 👍

PR #193, opened 2018-11-27
Add heading anchor

### LINKED ISSUE #60: Add anchor link buttons next to section headings
Makes it easier to copy a link to a specific section

hanzei:
Pretty much taken from https://discourse.gohugo.io/t/adding-anchor-next-to-headers/1726/14

Screenshot:
screenshot from 2018-11-27 14-42-27

Fixes #60

### DIFF
--- a/site/layouts/_default/list.html
+++ b/site/layouts/_default/list.html
@@ -15,7 +15,7 @@
             <div class="row">
                 {{ partial "sidebar.html" .}}
                 <div class="col-lg-9 doc-content">
-                    {{ .Content }}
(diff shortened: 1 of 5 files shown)


hanzei:
> If we'd use docs.mattermost.com as an example, the height of the anchor would match the height of lower-case letters, and there would be one whitespace separating the heading and the anchor.

Just to get the facts strait: docs.mattermost.com uses an absolute height for the anchor. See 
screenshot from 2018-12-03 17-36-24


But :+1: for using a relative height. What do you thing about this?
New:
screenshot from 2018-12-03 18-00-38
Old:
image

PR #194, opened 2018-11-28
Improve Notification Bar

### LINKED ISSUE #5: Notification bar improvements
For the notification bar that can be enabled by setting params.notifcation.enable to true in config.toml
- [ ] Make it closable
- [ ] When at top, don't cover menus but float (stay at top of screen) on scroll

hanzei:
- [x] Make it closable
- [x] When at top, don't cover menus but float (stay at top of screen) on scroll

|  After |  Before |
|---|---|
| screenshot from 2018-11-28 16-14-45  |  screenshot from 2018-12-04 09-15-10  |

There a slight changes to the background. I'm not an CSS expert. For comparison  with a closed notification:

|  After |  Before |
|---|---|
| screenshot from 2018-11-28 16-11-13 | screenshot from 2018-11-28 16-11-18 | 

Fixes #5

### DIFF
--- a/site/static/css/styles.css
+++ b/site/static/css/styles.css
@@ -3415,7 +3415,7 @@ table>tbody>tr:nth-of-type(odd) {
     -webkit-transform: skew(0deg, -5deg);
     transform: skew(0deg, -5deg);
     background-image: linear-gradient(-135deg, #1ba4f7 0%, #4641ff 100%);
-    height: 90%;
+    height: 100%;
     left: 0;
     position: absolute;
     top: -15%;
@@ -3445,9 +3445,11 @@ table>tbody>tr:nth-of-type(odd) {
 .homepage header {
     background: transparent;
     color: #fff;
+    /*
     left: 0;
     position: absolute;
     top: 0;
+    */
     width: 100%;
 }
 
@@ -3479,7 +3481,8 @@ table>tbody>tr:nth-of-type(odd) {
 
 .homepage__intro {
     color: #fff;
-    padding: 14em 0;
+    padding-top: 8em;
+    padding-bottom: 14em;
     position: relative;
 }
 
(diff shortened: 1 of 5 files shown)


site/static/css/styles.css
```diff
@@ -3445,9 +3445,11 @@ table>tbody>tr:nth-of-type(odd) {
 .homepage header {
     background: transparent;
     color: #fff;
+    /*
     left: 0;
     position: absolute;
     top: 0;
+    */
```
jwilander: Left in on purpose or should we remove?

jwilander:
@jasonblais you might want to take a look. Note that the colour didn't change. We can change that to be more appropriate when we actually use the notification bar for something

site/static/css/styles.css
hanzei: This should be removed. Fixed.

jasonblais:
@hanzei @jwilander It's not immediately clear what the improvement is? (I couldn't decipher it from the diff file).

hanzei:
@jasonblais I updated the screenshots to better represent the changes. Please take a look.

jasonblais:
@hanzei Thank you! This helps a lot.

Can you help make the whitespace left and right of the text equal? Currently it's greater on the left hand side.

image


A designer suggested 32px on both sides, but I'm not sure if that's the appropriate value.

hanzei:
I changed it so the text has 32px spacing to the left edge and the close icon right. Is this what you meant?

screenshot from 2018-12-06 09-31-46

I looks unproportional to me because the text leans to the left side. I would rather have a equal spacing to the left and right edge.

jasonblais:
@hanzei Hmm, that feels wrong, I agree. 

I hate to do this, but can you help revert back to what you had and we'll merge it instead? I'm looking at the notification bar we use for the Mattermost server and the spacing matches with what you had before.

image

I really really apologize for making you go back and forth, I'll do a more careful job next time before suggesting a change :)

hanzei:
Sure thing, no problem. Reverting a commit isn't a big deal. :wink:

jasonblais:
> Sure thing, no problem. Reverting a commit isn't a big deal. 😉

Fair enough :)

PR #203, opened 2018-12-06
Update Editorconfig

hanzei:
If you are fine with this, I would apply this style to all files.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,8 +5,17 @@ root = true
 [*]
 end_of_line = lf
 insert_final_newline = true
+trim_trailing_whitespace = true
 charset = utf-8
 
-[*.md]
+[*.{md,css,html}]
 indent_style = space
 indent_size = 4
+
+[*.toml]
+indent_style = space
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -5,8 +5,17 @@ root = true
...
 indent_style = space
 indent_size = 4
+
+[*.toml]
+indent_style = space
+indent_size = 2
+
+[*.go]
+indent_style = tab
+indent_size = 2
```
hmhealey: Is the indent_size needed here? We don't have it defined in the server's .editorconfig, so I don't know if it does anything here.

hanzei: BTH it might not be needed there. But it does no harm here.

hanzei:
@hmhealey I would like to apply this style to all files. Are you fine with this?

.editorconfig
hmhealey: That's fine then. I'll just add Joram as a second reviewer then so we can make sure to get two reviews before merging this

PR #216, opened 2019-01-08
update vault documentation

cpanato:


amyblais:
@cpanato Is this WIP or ready for review?

cpanato:
yes @amyblais

jwilander:
I will try the new instructions today

jwilander:
Tested and it worked for me. @cpanato can you give me a quick summary as to the reason for the update?

cpanato:
There is a bug in the sshd which vault is using and the workaround for that is generating this `ed25519` which is more security as well (as bonus). and then we need to update this doc to reflect the new usage

PR #224, opened 2019-01-25
add dev setup for ubuntu 18.04

cpanato:
Add Dev setup steps for the new ubuntu 18.04


jwilander:
@cpanato good for review?

cpanato:
@jwilander yes

hanzei:
@cpanato From what I can see the only difference to the Ubuntu 16.04 guide is the use of `apt` instead of `apt-get`. But `apt-get` is available on Ubuntu 18.04.

Are there other differences? Why do we need this guide?

cpanato:
@hanzei yes, i was thinking about that and maybe is better just update the tab name to add `Ubuntu 16.04/18.04`
thanks for the feedback

cpanato:
done @hanzei

PR #233, opened 2019-02-08
Minor formatting update for internal/writing-a-blog-post.md, so that numbered list appears correctly

jasonblais:
Observed:
image


jasonblais:
Hm, that didn't resolve it, so I just created an issue for now https://github.com/mattermost/mattermost-developer-documentation/issues/234

PR #239, opened 2019-02-16
Update push proxy instructions

enahum:
Should be merged after https://github.com/mattermost/mattermost-push-proxy/pull/38


hmhealey:
Did you mean to remove ` site/static/img/proxy-config.png`? It looks like it's still used a few lines below your other changes

enahum:
@hmhealey This is weird, I did not remove the image, I did replace it with another one.. so probably is showing the difference in bytes, not that the file was deleted.

hmhealey:
Oh yeah, I see the difference now. It's -92% in size as opposed to being removed completely

PR #257, opened 2019-03-16
fix wrong cd path at step 8

svelle:
the `cd` command at step 8 was missing the `src/github.com/` resulting in an error when using the command

### DIFF
--- a/site/content/contribute/server/developer-setup/windows.md
+++ b/site/content/contribute/server/developer-setup/windows.md
@@ -57,7 +57,7 @@
 8. Start the server and test your environment:
 
     ```sh
-    cd $(go env GOPATH)/mattermost/mattermost-server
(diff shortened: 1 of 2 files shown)


site/content/contribute/server/developer-setup/windows.md
```diff
@@ -57,7 +57,7 @@
 8. Start the server and test your environment:
 
     ```sh
-    cd $(go env GOPATH)/mattermost/mattermost-server
+    cd $(go env GOPATH)src/github.com/mattermost/mattermost-server
```
hanzei: Should this be
```suggestion
    cd $(go env GOPATH)/src/github.com/mattermost/mattermost-server
```

?

svelle:
@hanzei done

svelle:
> Does this only affect Windows?

yes, other paths are correct.

windows has some more problems to fix but I haven't found the time to take care of those yet.

PR #273, opened 2019-04-28
Update developer documentation

ali-farooq0:
When I set up the mattermost-server/webapp repos locally, I couldn't figure out who to create a new user to log into mattermost with. Adding two commands to the developer flow on how to create a user on initial setup.


jwilander:
Thanks for the addition @ali-farooq0 !

PR #279, opened 2019-05-02
Added Christpher Poile to Core Committers and R&D

levb:


site/content/internal/rd-teams.md
```diff
@@ -34,12 +34,13 @@ weight: 110
 
 ## Extensibility Team
 
+* Jason Blais - PM
```
jasonblais: Other teams seem to be ordered by 
 - Eng Lead
 - Engineer
 - PM
 - QA
 - UX

0/5, maybe keep same order for consistency with other teams?

lieut-data: And within the engineers, I think we order by hire date.

PR #283, opened 2019-05-09
[MM-15465] Add channel_id and type to curl request to avoid crash when push notification is pressed

migbot:


enahum:
@migbot it happens automatically after some time

PR #285, opened 2019-05-14
update plugins

cpanato:


site/content/internal/infrastructure/plugins.md
```diff
@@ -1,19 +1,19 @@
...
+subsection: Infrastructure
 weight: 40
 ---
 
-To add a plugin to community.mattermost.com, you need to do the following:
+To add a plugin to https://community.mattermost.com, you need to do the following:
 
 1. Add the configuration for your plugin and enable it by modifying PluginSettings and PluginState here: https://github.com/mattermost/platform-private/blob/master/kubernetes/community-kubernetes/configmap-config.yaml
-  - Copy those changes to https://github.com/mattermost/platform-private/blob/master/kubernetes/community-daily-kubernetes/configmap-config.yaml for community-daily
+  - community-daily is using the configuration in the Database, so to configure the plugin you need first deploy the plugin and then access the system console and configure the plugin and enable it.
```
levb: Some plugins may not function properly without required configuration. This should not happen, and we should guard against it, but at this point it happens often. Do we have a mechanism to pre-populate plugin config in the DB prior to installing? (even if ugly)

cpanato: ihmo if you add a plugin in the list to be installed it should be enabled, otherwise dont add in the list.

we dont have any mechanism to config the plugin in the db, i'm happy to have that

PR #294, opened 2019-05-30
Adding debugging instructions to plugin docs

carmo-evan:


hanzei:
@cpoile Adding you as reviewer because you did some work with delve and plugins before.

crspeller:
@carmo-evan You might want to add you should let the plugin continue executing soon after connecting. Otherwise the MM server will detect and attempt to restart it.

carmo-evan:
> @carmo-evan You might want to add you should let the plugin continue executing soon after connecting. Otherwise the MM server will detect and attempt to restart it.

Made the change. Please review.

PR #296, opened 2019-06-03
Update Cocoapods version to 1.6.1

hmhealey:
Bumping the minimum version number for Cocoapods since it keeps getting modified whenever I build the mobile app.


enahum:
@hmhealey I think I merge this too early, can we make a PR to the master branch just to update cocoapods?

PR #303, opened 2019-06-12
Changed github username.

skiplist:


hanzei:
Out of interest: Why did you change your username?

skiplist:
@hanzei I have had first-letter.last-name for 10 years and wanted a change. Also I was curious what would break.

PR #312, opened 2019-06-27
MM-15324 override ephemeral post example

scottleedavis:
#### Summary
Include example for registering a custom ephemeral post type

#### Ticket Link
* https://github.com/mattermost/mattermost-server/issues/10797

#### Related Pull Requests
* https://github.com/mattermost/mattermost-webapp/pull/2759
* https://github.com/mattermost/mattermost-webapp/pull/3024

### DIFF
--- a/site/content/extend/plugins/webapp/reference.md
+++ b/site/content/extend/plugins/webapp/reference.md
@@ -41,6 +41,7 @@ The entry point `index.js` of your application might contain:
 ```javascript
 import UserPopularity from './components/profile_popover/user_popularity';
 import SomePost from './components/some_post';
+import SomeEphemeralPost from './components/some_ephemeral_post';
 import MenuIcon from './components/menu_icon';
 import {openExampleModal} from './actions';
 
@@ -52,6 +53,10 @@ class PluginClass {
         registry.registerPostTypeComponent(
             'custom_somepost',
             SomePost,
+        );
+        registry.registerPostTypeComponent(
+            'custom_ephemeralpost',
+            SomeEphemeralPost,
         );
         registry.registerMainMenuAction(
             'Plugin Menu Item',


marianunez:
Does adding this ephemeral override post example here needed? I feel that with the basic post override example and what was added with this [PR](https://github.com/mattermost/mattermost-webapp/pull/3024) that is reflected in the detailed Registry section of this same document would be enough. 1/5 
Thoughts? @lieut-data

site/content/extend/plugins/webapp/reference.md
```diff
@@ -53,6 +54,10 @@ class PluginClass {
             'custom_somepost',
             SomePost,
         );
+        registry.registerPostTypeComponent(
```
lieut-data: Can we modify the server-side of this component to exercise this custom post type?

scottleedavis: Sounds good, and if I understand you correctly, I went to look for the server side example in documentation (first looked [server reference](https://developers.mattermost.com/extend/plugins/server/reference/), however it didn't seem appropriate.   Should it be an example at the top of the [GoDoc](https://godoc.org/github.com/mattermost/mattermost-server/plugin)?  Apologies if I have misunderstood.  (also, for QA & demo plugin possible testing there is the [/ephemeral_override](https://github.com/mattermost/mattermost-plugin-demo/pull/40))

lieut-data: Please disregard my comment. It seems I reviewed this code thinking it was a plugin, but it's obvious to me now we're in the documentation block. No idea what I was thinking! :D

That being said, I think I agree with @marianunez that this example might not be necessary /here/ -- naturally, we don't demonstrate all the possible calls in this place, but defer to the demo plugin.

@scottleedavis, thoughts?

lieut-data: (To complete the chain, I think https://github.com/mattermost/mattermost-webapp/pull/3024 was basically sufficient.)

scottleedavis: @lieut-data I agree with both of you that this example is not needed.  Thank you for your feedback and time.

scottleedavis:
Thank you both again for your time.

PR #313, opened 2019-07-01
MM-16557 Add homepage link to admin docs 

aaronrothschild:
Add new homepage link to the admin docs, so developers who have setup their environments can understand how to login to Mattermost, then use it properly as an administrator.

As per JIRA https://mattermost.atlassian.net/browse/MM-16557

### DIFF
--- a/site/layouts/partials/head.html
+++ b/site/layouts/partials/head.html
@@ -35,3 +35,29 @@
     gtag('js', new Date());
     gtag('config', 'UA-64458817-2');
 </script>
+
+<!-- Marketo Munchkin Tracking code for Marketing Team -->
+<script type="text/javascript">
+    (function() {
+    var didInit = false;
+    function initMunchkin() {
+    if(didInit === false) {
+    didInit = true;
+    Munchkin.init('161-FBE-733');
+    }
+    }
+    var s = document.createElement('script');
+    s.type = 'text/javascript';
+    s.async = true;
+    s.src = '//munchkin.marketo.net/munchkin.js';
+    s.onreadystatechange = function() {
+    if (this.readyState == 'complete' || this.readyState == 'loaded') {
+    initMunchkin();
+    }
+    };
+    s.onload = initMunchkin;
+    document.getElementsByTagName('head')[0].appendChild(s);
+    })();
+</script>
+
+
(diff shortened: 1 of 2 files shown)


amyblais:
Thank you!

aaronrothschild:
@amyblais  Sorry, another change came about....we had to add Marketo tracking scripts to the site.  I've placed the script in the <head> section.

site/layouts/partials/head.html
```diff
@@ -35,3 +35,29 @@
...
+    s.src = '//munchkin.marketo.net/munchkin.js';
+    s.onreadystatechange = function() {
+    if (this.readyState == 'complete' || this.readyState == 'loaded') {
+    initMunchkin();
+    }
+    };
+    s.onload = initMunchkin;
+    document.getElementsByTagName('head')[0].appendChild(s);
+    })();
+</script>
```
jwilander: @aaronrothschild can you give me a brief overview of what this is doing? I want to be clear on the data we're tracking from visitors to the developer site

aaronrothschild: From Marketo's site:

The default behavior of Marketo Munchkin is to do the following on page load:

1. Check to see if the current browser has a Munchkin cookie and create one if it is not there
2. Send a “Visit Web Page” event to the designated Marketo instance using the information from the current page and browser.  This records an activity to the corresponding record in Marketo.
3. Send “Clicked Link on Web Page” event for any user clicks that occur on links.

On a lead’s first visit to a page on your domain, a new anonymous lead record will be created in Marketo.  The primary key for this record is the Munchkin cookie (_mkto_trk) which is created in the user’s browser.  All subsequent web activity on that browser will be recorded against this anonymous record.  In order to be associated to a known record in Marketo, one of the following things must occur:

- The lead must visit a Munchkin-tracked page with an mkt_tok parameter in the querysytring from a tracked Marketo email link.
- The lead must fill out a Marketo Form.
- A Munchkin associateLead call must be sent to your Marketo instance.
- A SOAP syncLead or REST Associate Lead call must be sent.
- Once one of these conditions is fulfilled, the cookie and all associated web activity will be associated to the known lead.

jwilander: Thanks Aaron. Looks like it's mainly just tracking what they've clicked on/visited so that's fine with me

aaronrothschild:
@amyblais I added you back, but there are no visible "editorial" changes to users for you to check...just added tracking scripts for marketing team to the same home page.

PR #316, opened 2019-07-05
Remove /usr/local/go before installing over an existing version

noxer:
Some developers have got older versions installed already. This step helps to avoid multiple Go installations in the same directory.

### DIFF
--- a/site/content/contribute/server/developer-setup/ubuntu.md
+++ b/site/content/contribute/server/developer-setup/ubuntu.md
@@ -24,6 +24,7 @@
     ```sh
     sudo apt-get install -y build-essential
     wget https://storage.googleapis.com/golang/go1.12.linux-amd64.tar.gz
+    sudo rm -rf /usr/local/go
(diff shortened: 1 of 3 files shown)


site/content/contribute/server/developer-setup/ubuntu.md
```diff
@@ -24,6 +24,7 @@
     ```sh
     sudo apt-get install -y build-essential
     wget https://storage.googleapis.com/golang/go1.12.linux-amd64.tar.gz
+    sudo rm -rf /usr/local/go
```
lieut-data: Can you replicate this to `arch.md` and `centos.md`?

noxer: Sure.

noxer: The command only makes sense on CentOS, Arch uses pacman which takes care of the check.

noxer:
@lieut-data thank you for the feedback. I'll merge in the future.

PR #320, opened 2019-07-06
Remove useless link from developers.mattermost.com

aaronrothschild:
I'm not sure why all three commits of mine are here in this one branch, but I will assume this will all get squashed and merged together. @amyblais - should I be doing something differently?

- Add Marketo Code to header on all pages
- Add `Admin Docs` link to navbar of site
- Remove a useless link from the onboarding page

### DIFF
--- a/site/config.toml
+++ b/site/config.toml
@@ -81,6 +81,12 @@ pygmentsStyle = "manni"
       name = "Blog"
       weight = 4
 
+    [[menu.postpend]]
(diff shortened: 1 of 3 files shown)


amyblais:
Not sure, Joram do you know the answer to the above question?

jwilander:
@aaronrothschild you need to update your fork and branch with latest master. That's why the commits are still showing up

PR #325, opened 2019-07-20
update the documentation

Selimix:
Add the comment for the env var


site/content/contribute/mobile/build-your-own/assets-white-label.md
```diff
@@ -11,6 +11,9 @@ If you look at the [Project Directory Structure](/contribute/mobile/developer-wo
 
 To replace these with your own assets, create a sub-directory called `override` in the `assets` folder. Using the same directory structure and file names as in the `base` directory, you can add assets to the override folder to be used instead.
 
+In order to use the news assets, you need to change up the [Environment Variables](contribute/mobile/build-your-own/environment-vars/)  REPLACE_ASSETS to True (default is false)
```
enahum: this should be in the `App Splash Screen and Launch Icons` section as it is specific for that and does not apply for the in-app assets.

the value should be `true` and not `True`

Selimix: yes let me fix

PR #334, opened 2019-07-22
Remove unnecessary word in sentence.

jfrerich:
#### Summary 

removed unnecessary "we" in sentence.


jfrerich:
@hmhealey This documentation is and was really helpful during onboarding.  I made some notes when I first started that are very much `nits`, but figure I'll submit them now :)

PR #335, opened 2019-07-22
Add step to have a dedicated instance as part of community plugins

jasonblais:
Included in QA steps as it could be part of testing, but may also fall under PM or Dev ownership

Came up when Chris S tried to test Skype4Business plugin

### DIFF
--- a/site/content/extend/plugins/community_process.md
+++ b/site/content/extend/plugins/community_process.md
@@ -58,7 +58,7 @@ QA pass involves getting a memeber of our QA team to take a look and verify the 
 
 Steps involved:
 - Ensure all setup documentation needed is clear and can be successfully followed 
-- Dedicated instance for a third-party service is deployed, if applicable
+- Dedicated instance or test account to access and test the third-party service, if applicable
 - Functional testing has been done to ensure the integration works as expected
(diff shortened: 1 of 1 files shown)


site/content/extend/plugins/community_process.md
```diff
@@ -58,6 +58,7 @@ QA pass involves getting a memeber of our QA team to take a look and verify the
 
 Steps involved:
 - Ensure all setup documentation needed is clear and can be successfully followed 
+- Dedicated instance for a third-party service is deployed, if applicable
```
aaronrothschild: I'm not sure I understand....what's an example?  Like a test account for Service X?

jasonblais: An example is an Azure or Microsoft account with a running on-prem Skype4Business server instance to review and test Skype4Business plugin. I tested the cloud version with a one-off set up for the bounty, but if we want to support it officially, it'll be more efficient to set up an instance that we can continuously test on.

aaronrothschild:
```suggestion
- Dedicated instance or test account to access and test the third-party service, if applicable
```

PR #336, opened 2019-07-24
Require `2: QA Review` before merging

lieut-data:
As discussed in today's Developer Meeting, we're going to block all changes on a `2: QA Review` step, done concurrently with `2: Dev Review`. Not every pull request requires an exhaustive QA review, but it is the PM or QA tester's responsibility to determine this.

(Labels in GitHub to be updated once this PR is approved.)


jwilander:
I updated the labels on the repos

PR #344, opened 2019-08-05
Bump hugo version

hanzei:
This PR:
- Bumps the Hugo version for the build
- Fixes problems with the `note` shortcode with Hugo >0.54
- Fixes some depreciations and warnings

Please note that the code is still compatible with older Hugo versions.

### DIFF
--- a/site/content/contribute/mobile/push-notifications/service.md
+++ b/site/content/contribute/mobile/push-notifications/service.md
@@ -124,32 +124,29 @@ and running the service under a `mattermost-push-proxy` user account with limite
 ### Test the Mattermost Push Notification Service
 * Verify that the server is functioning normally and test the push notification using curl:
  `curl http://127.0.0.1:8066/api/v1/send_push -X POST -H "Content-Type: application/json" -d '{"type": "message", "message": "test", "badge": 1, "platform": "PLATFORM", "server_id": "MATTERMOST_DIAG_ID", "device_id": "DEVICE_ID", "channel_id": "CHANNEL_ID"}'`
-
-    * Replace `MATTERMOST_DIAG_ID` with the value found by running the SQL query:
-    ```sql
-    SELECT * FROM Systems WHERE Name = 'DiagnosticId';
-    ```
-    * Replace `DEVICE_ID` with your device ID, which can be found using (where `your_email@example.com` is the email address of the user you are logged in as):
-    ```sql
-    SELECT
-       Email, DeviceId
-    FROM
-       Sessions,
-       Users
-    WHERE
-       Sessions.UserId = Users.Id
-          AND DeviceId != ''
-          AND Email = 'your_email@example.com';
-    ```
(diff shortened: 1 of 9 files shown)


site/content/contribute/mobile/push-notifications/service.md
```diff
@@ -124,32 +124,29 @@ and running the service under a `mattermost-push-proxy` user account with limite
...
-    * Replace `CHANNEL_ID` with the Town Square channel ID, which can be found using:
-    ```sql
-    SELECT Id FROM Channels WHERE DisplayName = 'Town Square';
-    ```
-    
-    {{% note "Migration" %}}
-    Remove the `apple:`, `apple_rn`, `android:` or `android_rn:` prefix from your device ID before replacing `DEVICE_ID`. Use that prefix as the `PLATFORM` (make sure to remove the ":").
-    {{% /note %}}
-
+  * Replace `MATTERMOST_DIAG_ID` with the value found by running the SQL query:
```
hanzei: This was badly aligned, which caused the code block to not render.

site/layouts/shortcodes/note.html
```diff
@@ -1,6 +1,3 @@
 <aside class="note">
-    <div class="note-icon">
-        {{partial "svg/exclamation.svg" (dict "size" "20px" ) }}
```
hanzei: This partial is just an empty svg file. I removed it because it would render as <p></p>`.

site/layouts/shortcodes/note.html
```diff
@@ -1,6 +1,3 @@
...
-    </div>
-    <div class="note-content">{{- .Inner -}}</div>
+    <div class="note-content">{{ .Inner | markdownify }}</div>
```
hanzei: This is needed because of breaking changes in https://gohugo.io/news/0.55.0-relnotes/

site/layouts/shortcodes/note.html
```diff
@@ -1,6 +1,6 @@
 <aside class="note">
-    <div class="note-icon">
-        {{partial "svg/exclamation.svg" (dict "size" "20px" ) }}
+    <div class="note-content">
+        <strong>{{ .Get 0 }}</strong><br>
+        {{ .Inner | markdownify }}
```
hanzei: This is needed because of breaking changes in https://gohugo.io/news/0.55.0-relnotes/

site/layouts/shortcodes/note.html
```diff
@@ -1,6 +1,6 @@
 <aside class="note">
-    <div class="note-icon">
-        {{partial "svg/exclamation.svg" (dict "size" "20px" ) }}
+    <div class="note-content">
+        <strong>{{ .Get 0 }}</strong><br>
```
hanzei: We used the shortcode in the past like this parameter had any affect. This just fixes the behaviours.

hanzei:
@lieut-data I'm not sure who should review this. What do you think?

PR #370, opened 2019-09-27
Update code-review.md

lindy65:
Made changes to reflect the new "QA Review Done" label

### DIFF
--- a/site/content/contribute/getting-started/code-review.md
+++ b/site/content/contribute/getting-started/code-review.md
@@ -54,7 +54,7 @@ If you are a core committer seeking a review
 4. Assign a [QA tester](/contribute/getting-started/core-committers/#qa-testers) and label your pull request with `2: QA Review`.
     * It is the PM or QA tester's responsibility to determine the scope of required testing, if any.
     * Reviews by QA may occur at the same time as review by core committers. Be sure to ask for a second review as needed when changes are made.
-    * The QA tester should remove `2: QA Review` when their review is done, or if a review is deemed not necessary.
+    * The QA tester should add `QA Review Done` when their review is done, or if a review is deemed not necessary.
 5. Apply additional labels as necessary:
     * `CherryPick/Approved`: Apply this if the pull request is meant for a quality or patch release.
     * `Do Not Merge/Awaiting PR`: Apply this if the pull request depends on another (e.g. server changes)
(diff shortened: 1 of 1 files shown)


site/content/contribute/getting-started/code-review.md
```diff
@@ -54,7 +54,7 @@ If you are a core committer seeking a review
 4. Assign a [QA tester](/contribute/getting-started/core-committers/#qa-testers) and label your pull request with `2: QA Review`.
     * It is the PM or QA tester's responsibility to determine the scope of required testing, if any.
     * Reviews by QA may occur at the same time as review by core committers. Be sure to ask for a second review as needed when changes are made.
-    * The QA tester should remove `2: QA Review` when their review is done, or if a review is deemed not necessary.
+    * The QA tester should add `QA Review Done` when their review is done, or if a review is deemed not necessary.
```
saturninoabril: We'll still have to remove `2: QA Review`, right?

lindy65: Oh my gosh! Yes, of course :) Let me fix these...

lindy65: Fixed!

hanzei:
I'm wondering why this change was made. Before this, a new PR would get the labels `2. Dev Review` and `2. QA Review`. This makes clear, that these two review types are needed. What is the value that `QA Review Done` adds?

saturninoabril:
@hanzei Sorry for not being clear on this.  Main objective is to help quickly distinguish between the state of the QA Review label having been removed vs. not ever added.  Another is for reporting which we're currently working on to measure the impact of QA reviews in our process (request volume vs available mana in QA team).
I was not in the recent R&D meeting but I thought it was announced there as well.

cc @lindalumitchell

hanzei:
> Main objective is to help quickly distinguish between the state of the QA Review label having been removed vs. not ever added.

I 100% see that point of this. But I'm wondering if this is the right way to fix this. We could just add the `2. QA Review` via automation by default. That would be less manual work.

lindy65:
Hi @hanzei - thanks for your feedback :)

If the `2: QA Review` label could be added automation by default, that would be fantastic - and I agree, less manual work.

However, this PR was to address the new process where QA adds the `QA Review Done` label once they have reviewed the PR. This helps the QA process whereby we know at a glance that a PR has been reviewed. Previously, before this label existed, we had to scan the PR comments to see whether the `2: QA Review` label had been added for review and then again removed after review.

I'd be happy to submit a new PR to update the documentation if adding of the `2: QA Review` becomes automated.

Hope this helps?

hanzei:
Thank you very much for the explanation. :+1: 

@cpanato Do you think we can accomplish this with Mattermod? Or do you think we should use GitHub issues?

lieut-data:
@hanzei, I love the idea of https://github.com/actions/labeler automatically applying `2: QA Review`! The only caveat is that /sometimes/ it doesn't make sense to engage QA while there's ongoing developer review. Our numbering scheme doesn't really suggest this, but it has come up a few times with the idea of making `3: QA Review` a thing.

hanzei:
I would like to continue the discussion about automating these kind of workflows, but I don't think this is the right place for this. Maybe we continue the discussion on the community server or in a meeting?

lindalumitchell:
Yes, there are discussions currently happening in both meetings and channels. :) For now can we just go with the changes as written, then update if/when the process is officially updated?

hanzei:
@lindalumitchell Surely :wink:

PR #374, opened 2019-09-30
document plugin helpers methodology

lieut-data:
This addition to the plugin documentation explains the purpose of plugin helpers and guides developers on when to extend using same.

### DIFF
--- a/site/content/extend/plugins/helpers.md
+++ b/site/content/extend/plugins/helpers.md
@@ -1,0 +1,84 @@
+---
+title: Plugin Helpers
+date: 2019-09-30T00:00:00-05:00
+subsection: Plugins (Beta)
+weight: 100
+---
+
+The plugin architecture consists of an [API](API), [hooks](http://localhost:1313/extend/plugins/server/reference/#Hooks), and plugin helpers.
+
+When a plugin invokes an API method, it makes an RPC call to the Mattermost server and waits for a response. When the Mattermost server invokes a hook method, it makes an RPC call to the plugin and waits for a response:
+
+```go
+// OnActivate is a hook called by the server when the plugin is started.
+func (p *Plugin) OnActivate() error {
+    // CreatePost is an API called by the plugin to create a post.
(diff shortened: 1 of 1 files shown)


site/content/extend/plugins/helpers.md
```diff
@@ -0,0 +1,84 @@
+---
+title: Plugin Helpers
+date: 2019-09-30T00:00:00-05:00
+subsection: Plugins (Beta)
+weight: 100
+---
+
+The plugin architecture consists of an [API](API), [hooks](http://localhost:1313/extend/plugins/server/reference/#Hooks), and plugin helpers.
```
amyblais: ``hooks`` link doesn't work.

site/content/extend/plugins/helpers.md
```diff
@@ -0,0 +1,84 @@
...
+	GetUsers(options *model.UserGetOptions) ([]*model.User, *model.AppError)
+```
+
+Old servers won't do anything with new, unrecognized fields, but also won't break if they are present. When adding new options that will only be supported by some Mattermost server versions, consider wrapping the functionality with a plugin helper to help plugin authors safely use the API.
+
+## How do I add a plugin helper?
+
+To add a helper to the official [github.com/mattermost/mattermost-server/plugin](https://godoc.org/github.com/mattermost/mattermost-server/plugin#Helpers) package:
+
+* Extend the [`Helpers` interface](https://github.com/mattermost/mattermost-server/blob/master/plugin/helpers.go) with the new method
```
amyblais:
```suggestion
* Extend the [`Helpers` interface](https://github.com/mattermost/mattermost-server/blob/master/plugin/helpers.go) with the new method.
```

site/content/extend/plugins/helpers.md
lieut-data: Oops! Thanks -- fixed :)

hanzei:
I'm wondering is this file should be in https://developers.mattermost.com/extend/plugins/server/ instead, because it's just server plugins thing.

PR #377, opened 2019-10-01
Document requirement to install docker-compose

### LINKED ISSUE #364: Request for Documentation: Document requirement to install docker-compose
Mattermost user `jesse.hallam` from https://community-release.mattermost.com has requested the following be documented:

```
`docker-compose` is now required for developers. While our setup guide alludes to this, we should explicitly ask the user to install `docker-compose`.
```

See the original post [here](https://community-release.mattermost.com/_redirect/pl/xjncudp6nfb4dprgyj6f1sqa9e).

_This issue was generated from [Mattermost](https://mattermost.com) using the [Doc Up](https://github.com/jwilander/mattermost-plugin-docup) plugin._

DropNib:
This adds a new step to the developer setup documenting how to install the latest docker-compose version. This should fix #364.

Since this step fits best right after the installation of Docker, I inserted it as second step and renumbered the steps which follow it.

The installation commands are taken directly from the [official documentation](https://docs.docker.com/compose/install), except that I changed the download URL to always download the latest version in order to avoid always having to update this documentation.
I only added instructions for platforms, which according to the manual on don't come already with a current docker-compose version:
- Arch
- Centos
- Ubuntu
- Windows WSL

For the other platforms, only the numbering was updated, to reflect the addition of a new, omitted step.

The one thing I wasn't sure on how to incorporate was the notice in the normal Docker installation step regarding the migration to docker-compose:
> **Note**: [MM-9791](https://github.com/mattermost/mattermost-server/pull/10872) introduced using [docker-compose](https://docs.docker.com/compose/) to manage containers. To preserve your data on upgrade, execute the following steps.
...

But after trying out different variations, leaving them as they are seems the best option to me.


site/content/contribute/server/developer-setup/osx.md
```diff
@@ -15,38 +15,38 @@
     If you don't migrate your data, the new, docker-compose-managed containers will start out empty. To remove the old containers -- destroying any existing data -- use `make clean-old-docker`.
      
     
-2. Download and install homebrew: https://brew.sh/.
+3. Download and install homebrew: https://brew.sh/.
```
lieut-data: Is the `brew install docker-compose` step missing here?

DropNib: @lieut-data Thanks for the review.

As stated in the [official documentation for docker-compose](https://docs.docker.com/compose/install):
> **Docker Desktop for Mac** and **Docker Toolbox** already include Compose along with other Docker apps, so Mac users do not need to install Compose separately.

I can confirm this on my Mac - I just installed Docker Desktop for Mac and docker-compose is already included:

```sh
$ docker-compose --version
docker-compose version 1.24.1, build 4667896b
```

site/content/contribute/server/developer-setup/osx.md
```diff
@@ -15,38 +15,38 @@
...
 
-3. Install Go:
+4. Install Go:
     ```sh
     brew install go
```
cpoile: @lieut-data While we're at it, should we:
```suggestion
     brew install go@1.12
```

site/content/contribute/server/developer-setup/arch.md
```diff
@@ -21,41 +21,49 @@
...
+
+4. Install Go:
 
     ```sh
     sudo pacman -S base-devel
     sudo pacman -S go
     ```
 
-4. Update your shell's initialization script (e.g. `.bashrc` or `.zshrc`) and add the following:
+5. Update your shell's initialization script (e.g. `.bashrc` or `.zshrc`) and add the following:
```
amyblais: I don't know if the list numbers go correctly on these, but they look fine on preview.

site/content/contribute/server/developer-setup/osx.md
lieut-data: Ah, perfect! Should revert the section increment then to keep this as the second step?

site/content/contribute/server/developer-setup/windows-wsl.md
```diff
@@ -22,8 +22,18 @@ This is an unofficial guide. Community testing, feedback and improvements are we
...
 
-4. Install Go (using bash):
+    ```sh
+    sudo curl -L "https://github.com/docker/compose/releases/latest/download/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
+    sudo chmod +x /usr/local/bin/docker-compose
+    ```
+
+4. Install docker-compose
+   
+   You should install docker-compose
```
amyblais:
```suggestion
   You should install docker-compose.
```

site/content/contribute/server/developer-setup/osx.md
lieut-data: I don't oppose the change, but I'd probably recommend our fixing that separately since I expect it's a bunch of research to determine how to install that version on other platforms.

cpoile: Looks like it's Arch and Windows (non wsl) that don't have 1.12 specific. I'm fine for a separate PR for it, it was just a thought.

site/content/contribute/server/developer-setup/arch.md
DropNib: The hugo server usually ignores the numbering of the source and just counts incrementally. So in the source you could number a list like this:

```md
12.   
999.    
1.   
```
and hugo would still output:

```md
1.
2.
3.
```

I only took the existing list, inserted my step and applied +1 on every following step. I think however, that the numbering in all the edited files is slightly off.

Maybe we could address that in a separate issue? Or I could add a commit renumbering all the steps in a linear sequence to this PR.

site/content/contribute/server/developer-setup/windows-wsl.md
DropNib: Wait a minute, the entire step 4 isn't supposed to be here - that was simply a draft which I replaced (or at least I thought so) later with the real version.
I added a new commit to this PR which removes this old artifact and properly numbers the docker-compose installation step.

Sorry for the inconvenience :/

site/content/contribute/server/developer-setup/osx.md
DropNib: I think I misinterpreted the numbering a bit. I saw that in every file except the `osx.md` step number 2 was missing (jumping straight from step 1 to 3) and step number 2 in `osx.md` was `Download and install homebrew`, which of course is only necessary for MacOS. So I thought that step 2 _always_ is meant to be `Download and install homebrew` and the absence of step 2 in all other files was to indicate that brew is not to be installed.

So then when I added my step which was necessary before installing brew, I thought I had to reflect that by incrementing the numbering of the list.

Since that adds confusion, I could either add a new commit reverting all those numbering changes or close this PR and submit a new one without those numbering changes.

lieut-data: I don't think there's any expectation of uniform numbering between the operating systems, so I'd recommend just pushing a new commit to revert that file. Thanks, @DropNib!

site/content/contribute/server/developer-setup/arch.md
```diff
@@ -21,41 +21,49 @@
     ```
     If you don't migrate your data, the new, docker-compose-managed containers will start out empty. To remove the old containers -- destroying any existing data -- use `make clean-old-docker`.
     
-3. Install Go:
+2. Install docker-compose
+
+    ```sh
+    sudo curl -L "https://github.com/docker/compose/releases/latest/download/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
+    sudo chmod +x /usr/local/bin/docker-compose
+    ```
```
hanzei: Instead of using this approach you can install it directly via `pacman`:
`sudo pacman -S docker-compose`

site/content/contribute/server/developer-setup/osx.md
DropNib: Alright, thanks @lieut-data. I pushed the reversion.

site/content/contribute/server/developer-setup/arch.md
DropNib: Ah yes, you're right. The reason that docker-compose is usually out of date in distro repos of course doesn't apply to Arch.

Fixed it in a new commit.

hanzei: Thanks :+1:

DropNib:
Thank you as well, @lieut-data and all the other reviewers. 😊

PR #382, opened 2019-10-03
Bump hugo version to 0.58.3

hanzei:
Bumping the hugo version to keep in in sync with the latest release. There have been problems in the past where we let the version get async to much which caused problems.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ sudo: false
 language: go
 go: "1.11"
 env:
-  - HUGO_VERSION="0.56.3"
+  - HUGO_VERSION="0.58.3"
 before_install:
   - mkdir bin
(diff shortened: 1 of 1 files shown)


hanzei:
@prapti Adding you as a reviewer for this as this change might have some unintended side effects. I think there is value in also QA reviewing these kinds of PRs.

This PR is totally not urgent. Feel free to defer the review until you have time.

prapti:
@hanzei Would love to look into it, although I'm not sure where to begin as I have no idea what this is. Would you mind putting down some test steps and/or testing areas (what and where I should be looking when verifying this and if there is any setup needed)? That'd be super helpful. And I'm sure, once I understand it better, it'll be much easier for us to tackle these. :)
Thank you!

PR #383, opened 2019-10-03
[MM-18676] Add tags to sort plugin server API methods

cinlloc:
#### Summary
- Modify scripts/plugin-godocs.go to add tags to JSON:
  - Each "@tag" found in method comments is added to new "Tags" array attribute for each method
  - A global "Tags" attribute is also appended at same level as "Methods"
- Modify HTML layout to handle "Tags" in server API method index
  - No impact if no "Tags" attribute is found

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-18676 aka https://github.com/mattermost/mattermost-server/issues/12265

#### Comments
Not so easy to do and test, as it touch several scripts that are not covered by tests. Would be glad to have your feedbacks! If we achieve something on this side, there will be another PR in `mattermost-server` to add correct tags to `api.go` methods. 
I did some choices that are not specified in ticket (or different) : 
* Replaced `@memberOf` by `@tag` which seems more consistent with feature description
* I only categorized index, because seems quite a bit messy with layout if we change the bottom part

### DIFF
--- a/scripts/plugin-godocs.go
+++ b/scripts/plugin-godocs.go
@@ -16,6 +16,8 @@ import (
 	"go/types"
 	"log"
 	"os"
+	"regexp"
+	"sort"
 )
 
 type Field struct {
@@ -25,6 +27,7 @@ type Field struct {
 
 type MethodDocs struct {
 	Name       string
+	Tags       []string `json:"Tags,omitempty"`
 	HTML       string
 	Parameters []*Field `json:"Parameters,omitempty"`
 	Results    []*Field `json:"Results,omitempty"`
@@ -32,6 +35,7 @@ type MethodDocs struct {
 
 type InterfaceDocs struct {
 	HTML    string
+	Tags    []string `json:"Tags,omitempty"`
 	Methods []*MethodDocs
 }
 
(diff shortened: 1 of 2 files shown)


scripts/plugin-godocs.go
```diff
@@ -53,6 +57,31 @@ func docHTML(text string) string {
 	return buf.String()
 }
 
+func removeDuplicates(array []string) []string {
+	keys := make(map[string]struct{})
+	set := []string{}
+	for _, element := range array {
+		if _, value := keys[element]; !value {
```
lieut-data: I'd suggest using the variable name `ok` instead of `value` to be more idiomatic.

scripts/plugin-godocs.go
```diff
@@ -53,6 +57,31 @@ func docHTML(text string) string {
...
+			set = append(set, element)
+		}
+	}
+	return set
+}
+
+func tags(doc string) []string {
+	tagRegexp := regexp.MustCompile(`@tag\s+(\w+)\s*`)
+	submatches := tagRegexp.FindAllStringSubmatch(doc, -1)
+	if (submatches == nil) {
```
lieut-data: nit: I think you can eliminate this check, and everything still works. 0/5 on whether the code is simpler as a result.

site/layouts/shortcodes/plugingodocs.html
```diff
@@ -102,6 +123,7 @@ <h3 id="Examples.{{ $name }}">(Example) {{ substr $name 1 1 | upper }}{{ substr
         <h2 id="API">API</h2>
         <p>{{ safeHTML $docs.API.HTML }}</p>
         {{ template "InterfaceTOC" dict "Name" "API" "Interface" $docs.API }}
+        {{ template "InterfaceTOCByTags" dict "Name" "API" "Interface" $docs.API }}
```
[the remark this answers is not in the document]
cinlloc: Agree

scripts/plugin-godocs.go
```diff
@@ -53,6 +57,31 @@ func docHTML(text string) string {
 	return buf.String()
 }
 
+func removeDuplicates(array []string) []string {
+	keys := make(map[string]struct{})
```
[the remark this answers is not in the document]
cinlloc: Ok

scripts/plugin-godocs.go
cinlloc: Ok

scripts/plugin-godocs.go
cinlloc: Ok

cinlloc:
Thanks for your comments. Here's a screenshot of the result. One can feel confusing that untagged methods are at same level as tags, but I haven't found another solution (except maybe adding an `<Untagged>` category at top level list). Any suggestion welcomed!

mattermost-plugin-api-doc-screenshot-1

cinlloc:
This server PR could help to do QA: https://github.com/mattermost/mattermost-server/pull/12701

PR #389, opened 2019-10-11
Fixed Javascript Link

hahmadia:
Javascript text was linking to golang website.


hanzei:
Thanks for the fix @hahmadia :+1:

PR #406, opened 2019-10-24
add github action to post when we have new blog post

cpanato:
This PR adds a github action to check when we have a new blog post and then post in the Developers channel in community

fixes jira: https://mattermost.atlassian.net/browse/IS-225

/cc @coreyhulen for visibility with what is an example that we can make with GH action and MM


.github/workflows/notify-mm-blog.yml
```diff
@@ -0,0 +1,32 @@
...
+          echo $FILEADDED
+          if [ -z "$FILEADDED" ]
+          then
+            echo "nothing added in the .md"
+            exit 0
+          else
+            FILENAME=$(cat $FILEADDED | grep slug | cut -d':' -f 2 | sed 's/^ *//g')
+            echo $FILENAME
+            TITLE=$(cat $FILEADDED | grep title | cut -d':' -f 2 | sed 's/^ *//g')
+            echo "{\"username\":\"GitHub Action Notify\",\"icon_url\":\"https://www.mattermost.org/wp-content/uploads/2016/04/icon.png\",\"attachments\":[{\"author_name\":\"mattermost.dev\",\"author_link\":\"https://mattermost.dev\",\"fallback\":\"New Blog Post! Check out [here](https://developers.mattermost.com/blog/)\",\"color\":\"good\",\"title\":\""$TITLE"\",\"text\":\"New Blog Post! Check out [here](https://developers.mattermost.com/blog/$FILENAME)\"}]}" > mattermost.json
```
lieut-data: Terrific! Just a minor prose suggestion:
```suggestion
            echo "{\"username\":\"GitHub Action Notify\",\"icon_url\":\"https://www.mattermost.org/wp-content/uploads/2016/04/icon.png\",\"attachments\":[{\"author_name\":\"mattermost.dev\",\"author_link\":\"https://mattermost.dev\",\"fallback\":\"New Blog Post! Check out [here](https://developers.mattermost.com/blog/)\",\"color\":\"good\",\"title\":\""$TITLE"\",\"text\":\"New Blog Post! Check it out [here](https://developers.mattermost.com/blog/$FILENAME).\"}]}" > mattermost.json
```

.github/workflows/notify-mm-blog.yml
```diff
@@ -0,0 +1,32 @@
...
+        run: |
+          set -x
+          git diff --name-only --diff-filter=A HEAD~1 -- 'site/content/blog/*.md' > file.log
+          FILEADDED=$(cat file.log)
+          echo $FILEADDED
+          if [ -z "$FILEADDED" ]
+          then
+            echo "nothing added in the .md"
+            exit 0
+          else
```
hanzei: Nit: We can drop the `else` here

.github/workflows/notify-mm-blog.yml
```diff
@@ -0,0 +1,32 @@
...
+            FILENAME=$(cat $FILEADDED | grep slug | cut -d':' -f 2 | sed 's/^ *//g')
+            echo $FILENAME
+            TITLE=$(cat $FILEADDED | grep title | cut -d':' -f 2 | sed 's/^ *//g')
+            echo "{\"username\":\"GitHub Action Notify\",\"icon_url\":\"https://www.mattermost.org/wp-content/uploads/2016/04/icon.png\",\"attachments\":[{\"author_name\":\"mattermost.dev\",\"author_link\":\"https://mattermost.dev\",\"fallback\":\"New Blog Post! Check out [here](https://developers.mattermost.com/blog/)\",\"color\":\"good\",\"title\":\""$TITLE"\",\"text\":\"New Blog Post! Check it out [here](https://developers.mattermost.com/blog/$FILENAME).\"}]}" > mattermost.json
```
hanzei: Can we use `author` from the post as `author_name`? And the `author_link` to the GH profile?

.github/workflows/notify-mm-blog.yml
```diff
@@ -0,0 +1,32 @@
+name: notify-mm-blog
+on:
+  push:
```
hanzei: Don't we want to limit this to `master`? Otherwise it will kick in on new files on branches, right?

```suggestion
  push:
    branches: 
      - master
```

cpanato: no, we want only in that path

.github/workflows/notify-mm-blog.yml
cpanato: not following

.github/workflows/notify-mm-blog.yml
cpanato: ok

.github/workflows/notify-mm-blog.yml
hanzei: Can't we do both? path and branch?

.github/workflows/notify-mm-blog.yml
hanzei: The md file as a `author` field in there front matter. I would like to use this field to set `author_name`. This way `author_name` will be e.g. "Ben Schumacher".

.github/workflows/notify-mm-blog.yml
cpanato: why?

.github/workflows/notify-mm-blog.yml
cpanato: done

.github/workflows/notify-mm-blog.yml
hanzei: Correct me if I'm wrong, but doesn't the current code post a message if we add a new file to `site/content/blog/*.md`? I think we only want to post it, when the file gets added to `master`, not when it get added to the feature branch.

cpanato: sounds fair

hanzei:
Are GitHub actions working now?

PR #419, opened 2019-11-05
document how to shut down and clean up docker containers

ashishbhate:
The developer docs mention `make stop-server` to stop the running server, but this does not stop the docker containers started by `make run-server`. 

Added a sentence about `make stop-docker` to stop and clean up the running docker containers.

### DIFF
--- a/site/content/contribute/server/developer-setup/arch.md
+++ b/site/content/contribute/server/developer-setup/arch.md
@@ -75,3 +75,9 @@
     ```
 
     **Note:** Browsing directly to http://localhost:8065/ will display a `404 Not Found` until the web app is configured. See [Web App Developer Setup](https://developers.mattermost.com/contribute/webapp/developer-setup/) and [Mobile App Developer Setup](https://developers.mattermost.com/contribute/mobile/developer-setup/) for additional setup.
+
+    The `stop-server` make target does not stop all the docker containers started by `run-server`. To stop and clean up running docker containers:
+
+    ```sh
+    make stop-docker
+    ```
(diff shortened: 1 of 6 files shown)


hanzei:
This is a follow up on https://community.mattermost.com/core/pl/yt76ebzymfbkf857ooeo1k3m1e

ashishbhate:
The make target for `start-docker` (which is called by `run-server`) passes the `--rm` switch to `docker-compose run` so I assumed that the containers were automatically cleaned up when `stop-docker` is called. 

You're right that the containers remain for reuse. I'll update the change here to remove the part of the clean up. The part about `clean-docker` and the destructive changes can be added in a separate pull request (if at all?). Does this work?

lieut-data:
I think that makes sense: thanks @ashishbhate!

ashishbhate:
I've updated the documentation to drop the part about the clean up.

site/content/contribute/server/developer-setup/arch.md
```diff
@@ -75,3 +75,9 @@
     ```
 
     **Note:** Browsing directly to http://localhost:8065/ will display a `404 Not Found` until the web app is configured. See [Web App Developer Setup](https://developers.mattermost.com/contribute/webapp/developer-setup/) and [Mobile App Developer Setup](https://developers.mattermost.com/contribute/mobile/developer-setup/) for additional setup.
+
+    The `stop-server` make target does not stop all the docker containers started by `run-server`. To stop the running docker containers:
+
+    ```sh
+    make stop-docker
```
cpanato: I did a PR that add the to the `make stop` the stop-docker step, i feel this was missing: https://github.com/mattermost/mattermost-server/pull/13020

PR #420, opened 2019-11-05
updated vault.md with instructions on loadtesting

reflog:
Added some `ltops` related notes

### DIFF
--- a/site/content/internal/infrastructure/vault.md
+++ b/site/content/internal/infrastructure/vault.md
@@ -73,3 +73,31 @@ When the Vault instance starts up, it needs to be "unsealed" in order to decrypt
 ```
 VAULT_ADDR='https://vault-sealed.mattermost.com' vault unseal
 ```
+
+## Accessing AWS Credentials via Vault
+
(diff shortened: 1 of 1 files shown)


reflog:
@DSchalla - why not both? Mostly for visibility...

DSchalla:
Mostly the concern that both needs to be maintained, and that this information is not particular useful to an outside contributor. The `internal` directory of the dev docs are in my opinion (this was discussed a few times in the past...) a bad split between "We don't want this to be public, but also not properly private" which we try to avoid. Change itself looks great and is a great addition.

reflog:
Ok. Will close this and add to the internal docs

DSchalla:
Thanks @reflog, appreciate it!

hanzei:
@DSchalla What is the reason that the `internal` docs still exist? Can't we just move everything to https://github.com/mattermost/internal-docs/?

DSchalla:
Absolutely. We are still migrating the content and adding additional documentation, but that's the goal overall.

PR #423, opened 2019-11-06
Update windows.md  change 'export' to 'set' for setting environment variables

allenlai18:
https://stackoverflow.com/questions/18701783/windows-equivalent-of-export

### DIFF
--- a/site/content/contribute/server/developer-setup/windows.md
+++ b/site/content/contribute/server/developer-setup/windows.md
@@ -25,12 +25,12 @@
 4. Setup the following environment variables (change the paths accordingly):
 
     ```sh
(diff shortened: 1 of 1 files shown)


allenlai18:
it should be 'set' in windows rather than 'export'   to set environment variables

allenlai18:
how do i rerun the check. i signed the CLA

hanzei:
@svelle Could you please review this?

svelle:
using export is technically correct here since we refer to using babun, however babun has been discontinued and we should provide an alternative for setting up the server environment on windows if we want to keep it in the documentation.

PR #446, opened 2019-11-28
Update the destination parameter for the git clone commands

jaydeland:
Adding a destination parameter so the clone commands sync the source to the default go path for the users system. Since I was pretty new at GoLang when I started this tripped me up some.


site/content/contribute/server/developer-setup/arch.md
```diff
@@ -57,13 +57,13 @@
 8. Clone the Mattermost source code from your fork:
 
     ```sh
-    git clone https://github.com/YOUR_GITHUB_USERNAME/mattermost-server.git
+    git clone https://github.com/YOUR_GITHUB_USERNAME/mattermost-server.git $(go env GOPATH)/src/github.com/mattermost/mattermost-server
```
hanzei: Since go 1.12 you don't need to put your code into `GOPATH`. Can you elaborate why this might be needed?

jaydeland:
I was trying to make it consistent with the mattermost-webapp developer documentation: https://developers.mattermost.com/contribute/webapp/developer-setup/

hanzei:
I think the webapp setup should just mention, that `server` and `webapp` have to be in the same directory. @jaydeland Would you be open on introducing this change?

jaydeland:
@hanzei Sure thing - at the sametime I think we can remove the manual step to create the client link as the Makefile is already doing that with the run-client target is run (or at least mention that it optional).

hanzei:
Awesome @jaydeland :+1:

PR #450, opened 2019-12-01
Switch to Circle CI pipeline

bpietraga:
## What it does and why

https://github.com/mattermost/mattermost-server/issues/13226

This PR provides `.circleci/config.yml` pipeline configuration with build stage which creates static webpage artifacts based on previous `travis.yml` configuration. 

Goals:
- Provide working static website deployment
- Align with Pipeline other repos using Circle CI to build and deploy
- Use Circle CI image with Go version `1.12.13`
- Remove `.travis.yml` configuration file.


cpanato:
@bpietraga can you please push an empty commit just to trigger circleci? i've enabled the project in the mattermost org

.circleci/config.yml
```diff
@@ -0,0 +1,75 @@
...
+          command: |
+            echo "export PATH=$GOPATH/bin:$PATH" >> $BASH_ENV
+            echo "export PATH=$PATH:$PWD/bin" >> $BASH_ENV
+            echo "HUGO_VERSION='0.58.3'" >> $BASH_ENV
+            echo "MATTERMOST_SERVER_RELEASE_TAG=v5.17.1" >> $BASH_ENV
+            source $BASH_ENV
+      - run:
+          name: install Node JS
+          command: |
+            sudo apt-get update
```
cpanato: you dont need to install node you can use the circleci image with node `circleci/golang:1.12-node`

.circleci/config.yml
```diff
@@ -0,0 +1,75 @@
...
+            sudo apt-get install -y nodejs npm
+      - run:
```
cpanato: checking the travis.yaml config i dont see all these steps there. why we need those?

.circleci/config.yml
bpietraga: Thanks, I overlooked this image. Fixed in the https://github.com/mattermost/mattermost-developer-documentation/pull/450/commits/d5fb2f2f062a2545cb4c8a43245f40b816b157c0

.circleci/config.yml
bpietraga: ### What Was Failing

Current `mattermost-server` master has breaking changes which cause script to fail.

https://travis-ci.org/mattermost/mattermost-developer-documentation/builds/620266705#L617

Based on the Travis CI output above the `scripts/plugin-godocs.go` fails after the
`mattermost-server` is installed in Makefile `backend-plugin-data`.
https://github.com/mattermost/mattermost-developer-documentation/blob/master/Makefile#L10

The breaking change to `mattermost-server` was introduced 5 days ago in the https://github.com/mattermost/mattermost-server/commit/2259b7f2a8b92b5dada097e77c284d5d14a01ba3

As the `mattermost-server` is not explicitly required in the `imports` of `scripts/plugin-godocs.go` , but during runtime this causes missing paths and failure of script. The example is in the Travis output.

I've tried to fix this issue using go mod `replace` directive to just point to  proper working release. So that we could use current `Makefile` with `make dist` like in the `travis.yml`. But this approach fails.
You can validate below.
The pipeline code would look like this:
https://github.com/mattermost/mattermost-developer-documentation/compare/master...bpietraga:test-2?expand=1
You can look failing pipeline for this `test-2` branch here:
https://circleci.com/gh/bpietraga/mattermost-developer-documentation/151

### Fix

I've solved this, using git checkout and rolling to release tag here:
https://github.com/mattermost/mattermost-developer-documentation/pull/450/files#diff-1d37e48f9ceff6d8030570cd36286a61R34

And then adding the steps from `Makefile`

I know that this is not ideal solution as the current `Makefile` commands fail.

cpanato: should we fix this in the source and not doing workarounds?

cpanato: will be fixed here: https://github.com/mattermost/mattermost-developer-documentation/pull/452

bpietraga: It is hard to not to agree with this one 😄 
Great! It is merged, so I've rebased the code with clean code. 
This time no workarounds. New commit: https://github.com/mattermost/mattermost-developer-documentation/pull/450/commits/e5870b7142a88d8cd7e4b91610efd67113f9f33f

cpanato: your solution was just an workaround and that will not work for the long run.

cpanato:
@metanerd @jaydeland please review when you have time, this is just moving from travis to cci no new things

hanzei:
FYI: The icon on https://github.com/mattermost/mattermost-developer-documentation/blob/master/README.md#L1 also needs to get updated. Should I file an issue or would someone just do this quickly?

cpanato:
you can do it @hanzei

PR #455, opened 2019-12-03
MM-19291 - Added documentation for custom plugin setting

marianunez:
Added documentation on how to create a custom plugin setting. I've included it as a new question within [Extend/Plugin/Best Practices](https://developers.mattermost.com/extend/plugins/best-practices/).

Ticket Link:

[MM-19291](https://mattermost.atlassian.net/browse/MM-19291)


site/content/extend/plugins/best-practices.md
```diff
@@ -6,6 +6,67 @@ weight: 50
 
 See here for [server-specific best practices for plugins](/extend/plugins/server/best-practices/). Webapp-specific best practices are incoming.
 
+## How can I create my own plugin setting?
+
+Plugins can define their admin settings in the plugin's manifest [setting schema](https://developers.mattermost.com/extend/plugins/manifest-reference/#settings_schema) which include a pre-defined types that the System Console UI currently supports.
```
lieut-data: I wonder if we should expand this section in order to emphasize that this is the "normal" way of configuring custom plugins and avoid someone diving into a custom component prematurely. Maybe even break it out as two questions, "How do I allow my plugin to be configured?" and "How do I handle more complex plugin configurations?"

site/content/extend/plugins/best-practices.md
```diff
@@ -6,6 +6,67 @@ weight: 50
...
 
+## How can I create my own plugin setting?
+
+Plugins can define their admin settings in the plugin's manifest [setting schema](https://developers.mattermost.com/extend/plugins/manifest-reference/#settings_schema) which include a pre-defined types that the System Console UI currently supports. 
+
+Additionally, a plugin could define their own type of setting with a corresponding custom user interface that will be displayed in the system console following the process below. 
+
+1. Define a setting in the plugins manifest setting scheme with a type of `custom`
+
+```
```
lieut-data: I find it useful to use the `diff` type so as to highlight the "changes". Perhaps something like:
```diff
"settings_schema": {
    "settings": [{
        "key": "NormalSetting",
        "type": "text",
+    }, {
+        "key": "CustomSetting",
+        "type": "custom"
    }]
}
```

site/content/extend/plugins/best-practices.md
justinegeffen: This gives a sense that the plugin is autonomous and defines its own schemas. The linked doc leans more toward the plugin being configurable by an Admin in the System Console. Pending confirmation, here's my suggestion:

```suggestion
A plugin's admin settings are defined in the plugin's manifest [setting schema](https://developers.mattermost.com/extend/plugins/manifest-reference/#settings_schema). This includes any pre-defined types that the System Console UI currently supports. 
```

site/content/extend/plugins/best-practices.md
```diff
@@ -6,6 +6,67 @@ weight: 50
 
 See here for [server-specific best practices for plugins](/extend/plugins/server/best-practices/). Webapp-specific best practices are incoming.
 
+## How can I create my own plugin setting?
+
+Plugins can define their admin settings in the plugin's manifest [setting schema](https://developers.mattermost.com/extend/plugins/manifest-reference/#settings_schema) which include a pre-defined types that the System Console UI currently supports. 
+
+Additionally, a plugin could define their own type of setting with a corresponding custom user interface that will be displayed in the system console following the process below.
```
justinegeffen: As per my previous comment, and here's my suggestion: 
```suggestion
Alternatively* a plugin can contain a custom setting, and a corresponding custom user interface will be displayed in the System Console. 
```
* Assuming the plugin can only be one or the other? Custom or pre-defined.

site/content/extend/plugins/best-practices.md
```diff
@@ -6,6 +6,67 @@ weight: 50
 
 See here for [server-specific best practices for plugins](/extend/plugins/server/best-practices/). Webapp-specific best practices are incoming.
 
+## How can I create my own plugin setting?
+
+Plugins can define their admin settings in the plugin's manifest [setting schema](https://developers.mattermost.com/extend/plugins/manifest-reference/#settings_schema) which include a pre-defined types that the System Console UI currently supports. 
+
+Additionally, a plugin could define their own type of setting with a corresponding custom user interface that will be displayed in the system console following the process below. 
+
+1. Define a setting in the plugins manifest setting scheme with a type of `custom`
```
justinegeffen:
```suggestion
1. Define a `type: custom` setting in the plugins manifest `settings_schema`
```

site/content/extend/plugins/best-practices.md
```diff
@@ -6,6 +6,67 @@ weight: 50
...
+
+```
+    "settings_schema": {
+        "settings": [{
+            "key": "CustomSetting",
+            "type": "custom"
+       }]
+```
+
+2. In the plugin Web App code, define a custom component to manage the plugin's specific setting and register in the Web App with [`registerAdminConsoleCustomSetting`](https://developers.mattermost.com/extend/plugins/webapp/reference/#registerAdminConsoleCustomSetting). This component will be instantiated in the System Console with the following `props` passed in:
```
justinegeffen:
```suggestion
2. In the plugin's Web App code, define a custom component to manage the plugin's custom setting and register in the Web App with [`registerAdminConsoleCustomSetting`](https://developers.mattermost.com/extend/plugins/webapp/reference/#registerAdminConsoleCustomSetting). This component will be represented in the System Console with the following `props` passed in:
```

site/content/extend/plugins/best-practices.md
```diff
@@ -6,6 +6,67 @@ weight: 50
...
+
+    - `id`: The setting `key` as defined in the plugin manifest within `settings_schema.settings`.
+    - `label`: The text for the component label based on the setting's `displayName` defined in the manifest. 
+    - `helpText`: The help text based on the setting's `helpText` defined in the manifest. 
+    - `value`: The setting's current json value in the config at the time the component is loaded.
+    - `disabled`: Boolean indicating if the setting is disabled by a parent component within the System Console.
+    - `config`: The server configuration loaded by the webApp.
```
justinegeffen:
```suggestion
    - `config`: The server configuration loaded by the WebApp.
```

site/content/extend/plugins/best-practices.md
```diff
@@ -6,6 +6,67 @@ weight: 50
...
+    - `license`: The license information for the related Mattermost server.
+    - `setByEnv`: Boolean that indicates if the setting is based on a server environment variable. 
+
```
justinegeffen: Remove space.

site/content/extend/plugins/best-practices.md
```diff
@@ -6,6 +6,67 @@ weight: 50
...
+    - `helpText`: The help text based on the setting's `helpText` defined in the manifest. 
+    - `value`: The setting's current json value in the config at the time the component is loaded.
+    - `disabled`: Boolean indicating if the setting is disabled by a parent component within the System Console.
+    - `config`: The server configuration loaded by the webApp.
+    - `license`: The license information for the related Mattermost server.
+    - `setByEnv`: Boolean that indicates if the setting is based on a server environment variable. 
+    
+    - `onChange`: Function that receives the setting id and current json value of the setting when it has been changed within the custom component. 
+    - `setSaveNeeded`: Function that will prompt the System Console to enable the Save button in the plugin settings screen. 
+
```
justinegeffen: Remove space.

site/content/extend/plugins/best-practices.md
```diff
@@ -6,6 +6,67 @@ weight: 50
...
+    - `registerSaveAction`: Registers the given function to be executed when the setting is being saved. This should be register when the custom component is mounted.
```
justinegeffen:
```suggestion
    - `registerSaveAction`: Registers the given function to be executed when the setting is saved. This is registered when the custom component is mounted.
```
Confirm that it is definitely registered or if there's a possibility for errors? If there's a possibility of errors then leave it as "should be" and consider a Troubleshooting section?

site/content/extend/plugins/best-practices.md
```diff
@@ -6,6 +6,67 @@ weight: 50
...
+    - `unRegisterSaveAction`: On unmount of the custom component, unRegisterSaveAction will remove the registered function executed on save of the custom component.
+
+3. On initialization of the custom component, the current value of the custom setting is passed in the `props.value` as a json read from the config. This value can be processed as necessary to display in your custom UI and ready to be modified by the end user. In the example below, it processes the initial `props.value` and sets it in a local state for the component to use as needed:
```
justinegeffen: Should it be "json value"?

site/content/extend/plugins/best-practices.md
```diff
@@ -6,6 +6,67 @@ weight: 50
...
+        super(props);
+
+        this.state = {
+            attributes: this.initAttributes(props.value),
+        }
+    }
+```
+
+
+4. Once the user makes any change in the UI, execute the `OnChange` handler to send back the current value of the setting as a json. Additionally, `setSaveNeeded` should be called to enable the `Save` button in order for the changes can be saved.
```
justinegeffen:
```suggestion
4. When a user makes a change in the UI, the `OnChange` handler sends back the current value of the setting as a json. Additionally, `setSaveNeeded` should be called to enable the `Save` button in order for the changes to be saved.
```

site/content/extend/plugins/best-practices.md
```diff
@@ -6,6 +6,67 @@ weight: 50
...
+```
+handleChange = () => {
+    ...
+
+    this.props.onChange(this.props.id,  Array.from(this.state.attributes.values()));
+    this.props.setSaveNeeded()
+};
+```
+
+5. Once the user clicks Save, any handler that was registered with `registerSaveAction` will be executed to perform any additional custom actions the plugin may require like calling an additional endpoint the plugin.
```
justinegeffen:
```suggestion
5. Once the user saves the changes, any handler that was registered with `registerSaveAction` will be executed to perform any additional custom actions the plugin may require, such as calling an additional endpoint the plugin. 
```

site/content/extend/plugins/best-practices.md
marianunez: A plugin can have both pre-defined and custom settings. I still believe your correction is valid: 

> Alternatively a plugin can contain a custom setting, and a corresponding custom user interface will be displayed in the System Console.

@justinegeffen Do you agree or would you change something more there?

site/content/extend/plugins/best-practices.md
marianunez: Could we keep `instantiated` here? Given that I'm referring to specific `props` that are passed at the time the component is created not sure if `represented` would have the same meaning.

site/content/extend/plugins/best-practices.md
marianunez: I wasn't clear enough here to indicate that this is an optional step where they can call `registerSaveAction` if they need to add extra processing after the setting.

site/content/extend/plugins/best-practices.md
marianunez: It comes in as a json but then you can work with it as needed, change it to whatever format you need in your component.

site/content/extend/plugins/best-practices.md
marianunez: @lieut-data are you suggesting adding a new question before 
"How do I allow my plugin to be configured?" and change this new content to "How do I handle more complex plugin configurations?"

site/content/extend/plugins/best-practices.md
marianunez: Updated 👍

marianunez:
> I've added some suggestions some of which include some questions regarding the content. The changes I suggested may not be in line with the intended content so review is needed.

Thanks @justinegeffen! I've included your suggestions and are up for a second proofread.

site/content/extend/plugins/best-practices.md
justinegeffen: I agree with that, I think it still works. :)

site/content/extend/plugins/best-practices.md
justinegeffen: Maybe "in json format"?

marianunez: Updated 👌

marianunez:
@aaronrothschild gentle ping to review so it can be available soon given this feature came out with v5.18.

aaronrothschild:
Thanks @marianunez - looks good - the documentation is really clear for implementing this.

site/content/extend/plugins/best-practices.md
marianunez: @lieut-data should we create an improvement ticket for this suggestion ?

marianunez: Ticket created - [MM-21364](https://mattermost.atlassian.net/browse/MM-21364)

PR #461, opened 2019-12-11
Update rd-teams.md

jaydeland:


lieut-data:
@jaydeland, you'll also want to update the core committers, I think: see https://github.com/mattermost/mattermost-developer-documentation/pull/462 for an example.

jaydeland:
> @jaydeland, you'll also want to update the core committers, I think: see #462 for an example.

@lieut-data - Looks like somebody already added us to the core committers page.

PR #494, opened 2020-01-14
Bump hugo version to 0.62.2

hanzei:
Bumping the hugo version to keep in in sync with the latest release. There have been problems in the past where we let the version get async to much which caused problems.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ jobs:
       - run:
           name: build Hugo
           environment:
-            HUGO_VERSION: 0.58.3
+            HUGO_VERSION: 0.62.2
           command: |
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -12,7 +12,7 @@ jobs:
       - run:
           name: build Hugo
           environment:
-            HUGO_VERSION: 0.58.3
+            HUGO_VERSION: 0.62.2
```
lieut-data: Do we need to do anything to ensure the local developer version is also up-to-date?

hanzei: Building the site doesn't _require_ the higher version. Hence, devs don't need to update.

But there might so some small parts like md rendering that are broken with an old version.

PR #505, opened 2020-01-17
Update developer-setup.md

Willyfrog:
#### Summary

Currently we are using `npm` instead of `yarn`


Willyfrog:
I probably wouldn't have noticed had we not the discussion today.

PR #507, opened 2020-01-18
Stopped the `):` from getting absorbed into the link

mroe:
#### Summary

This change delimits a link with less-than, greater-than symbols "< >" to avoid surrounding text from being absorbed into it.  In my tests,

```
3. Install Go (modify installation to latest Go version from https://golang.org/dl/):
```
effectively had a link at the end pointing to `https://golang.org/dl/):` instead of the intended `https://golang.org/dl/`. 

#### Ticket Link

None.  Minor fix.

### DIFF
--- a/site/content/contribute/server/developer-setup/ubuntu.md
+++ b/site/content/contribute/server/developer-setup/ubuntu.md
@@ -32,7 +32,7 @@
     sudo chmod +x /usr/local/bin/docker-compose
     ```
 
-3. Install Go (modify installation to latest Go version from https://golang.org/dl/):
+3. Install Go (modify installation to latest Go version from <https://golang.org/dl/>):
 
     ```sh
     sudo apt-get install -y build-essential


site/content/contribute/server/developer-setup/ubuntu.md
```diff
@@ -32,7 +32,7 @@
     sudo chmod +x /usr/local/bin/docker-compose
     ```
 
-3. Install Go (modify installation to latest Go version from https://golang.org/dl/):
+3. Install Go (modify installation to latest Go version from <https://golang.org/dl/>):
```
hanzei: What hugo version are you using? I tried your patch with Hugo v0.59.0. The output was the following HTML:
```html
<a href="https://golang.org/dl/):">https://golang.org/dl/):</a>
```

Maybe the link rendering was also changed with v0.60.0?

lieut-data: Ah, that makes sense -- @mitchellroe helped identify the v0.60.0 regression, and this likely occurred in that context. @hanzei, do we want to plan an upgrade to v0.60.0 soon, or just shelve that plan for now given ongoing discussion about the myriad sets of documentation tooling we have?

mitchellroe: I'm not sure what version I tested it under. I'll try it in hugo < 0.60.0 and in hugo >= 0.60.0 and see what happens.

hanzei: You fix works fine with Hugo >= 0.60 :+1:

Issue #506, opened 2020-01-18
contribute/server/developer-setup not rendering properly


mroe:
It looks like `contribute/server/developer-setup.md` is rendering correctly in my local Hugo installation, but fails on https://developers.mattermost.com/contribute/server/developer-setup/.

mitchellroe:
I've tested in Firefox 68.4.1esr and Google Chrome 79.0.3945.130 on Debian 10.

![image](https://user-images.githubusercontent.com/51406825/72670883-f30fad00-3a10-11ea-847a-100a15c34e44.png)

mitchellroe:
Here it is on my local Hugo installation:

![image](https://user-images.githubusercontent.com/51406825/72670909-4f72cc80-3a11-11ea-8c7f-748833306a7f.png)

mitchellroe:
I was running hugo v0.54.0/extended from the base Debian repositories, from which I made the above screenshot.  I upgraded my local hugo installation to v0.62.2-83E50184, and my experience is the same as on <https://developers.mattermost.com/contribute/server/developer-setup/>.

mitchellroe:
Looks like it broke with Hugo v0.60.0; v0.59.1 works fine.  I'll look through the release notes to see if there's something they changed which would cause this behavior.
https://github.com/gohugoio/hugo/releases/tag/v0.60.0

mitchellroe:
They switched from Blackfriday to Goldmark.
https://gohugo.io/news/0.60.0-relnotes/

Adding their recommended yaml block at the top of the file allows the HTML to be rendered.

```yaml
markup:
  goldmark:
    renderer:
      unsafe: true
```

hanzei:
Fixed via https://github.com/mattermost/mattermost-developer-documentation/pull/509

PR #508, opened 2020-01-19
Allow "unsafe" mode to use html in hugo source

mroe:
#### Summary

Hugo v0.60.0 switched from Blackfriday to Goldmark.  In this transition, they disabled some html tags in markdown files.  This allows html.

https://gohugo.io/news/0.60.0-relnotes/

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-developer-documentation/issues/506

### DIFF
--- a/site/content/contribute/server/developer-setup.md
+++ b/site/content/contribute/server/developer-setup.md
@@ -3,10 +3,6 @@ title: "Developer Setup"
 date: 2017-08-20T11:35:32-04:00
 weight: 2
 subsection: Server
-markup:
-  goldmark:
-    renderer:
-      unsafe: true
 ---
 
 <p>Set up your development environment for building, running, and testing the Mattermost server.</p>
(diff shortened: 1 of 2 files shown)


site/content/contribute/server/developer-setup.md
```diff
@@ -3,6 +3,10 @@ title: "Developer Setup"
 date: 2017-08-20T11:35:32-04:00
 weight: 2
 subsection: Server
+markup:
+  goldmark:
+    renderer:
+      unsafe: true
```
hanzei: Even with this change `contribute/server/developer-setup/` is still not rendered correctly at my local instance. I'm using `Hugo Static Site Generator v0.62.2/extended linux/amd64 BuildDate: unknown`

I also tried using `defaultMarkdownHandler: blackFriday` to just continue using black friday, but that did work either.

Do you have ides why this doesn't work on my local instance?

mitchellroe: Hmm.  I'm not sure.  I thought I tested it in 0.62.2, but I'm not sure.  I'll try it with that specific version and let you know what I find.

In the interim, it looks like the reversion of the version* seems to have worked.

* Potential metal band name

mitchellroe: Realized the configuration needs to be set in `config.toml`, not in the top of the file.  Pushed some changes.

PR #513, opened 2020-01-21
Bump hugo version to 0.62.2

hanzei:
#### Summary
Re submission of https://github.com/mattermost/mattermost-developer-documentation/pull/494.

Despite the issues that occurred when bumping the version the first time, I still consider it important to stay up to date with the latest Hugo version, because developers often update their local version and might encounter issues.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ jobs:
       - run:
           name: build Hugo
           environment:
-            HUGO_VERSION: 0.58.3
+            HUGO_VERSION: 0.62.2
           command: |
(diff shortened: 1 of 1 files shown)


lieut-data:
Thanks, @hanzei -- is there a way to see these changes live and review for potentially missed issues? I'm assuming we've already fixed the previous issues?

hanzei:
@lieut-data Unfortunately, since the move to CircleCI there is no preview via CD right now. I've filled https://mattermost.atlassian.net/browse/IS-314 to track this.

I've smoke tested the changes locally and didn't found any issues.

hanzei:
I will do more testing over the weekend.

PR #516, opened 2020-01-21
Use hugo-wrapper to unify version being used

khos2ow:
#### Summary

This PR adds [`hugow`](https://github.com/khos2ow/hugo-wrapper) which acts as a wrapper around Hugo binary which will have some benefits:

- the wrapper is platform independent and will download Hugo binary for the target system it runs on
- `.hugo/version` file can be used to control which version of Hugo should be used, this will unify the Hugo version in all systems (dev machines, CI, etc)

In reality this PR removes the need for #513 and controlling the version of Hugo from now on will be by changing `site/.hugo/version` file

/cc @hanzei


jaydeland:
This look interesting and I like the approach. I will review and would also like @metanerd to have a look.

hanzei:
Thanks @khos2ow for the PR.

I agree, that is an interesting approach. My main concern is, that is approach adds another moving part to the system with issues like:
- If there are bugs in `hugow`, we have to update the script
- People might "fork" the script and do manual updates
- I know the script is open source, but at least on my system, I would not want a script to install software.

I still leaning towards staying up to date with the latest hugo version. This can be even enforced by setting `min_version = "0.62.0"` in `theme.toml`.

lieut-data:
I wonder, too, if it might be more natural to define `hugo` relative to a Docker container, and use /that/ both to develop and build, ostensibly always keeping them in lockstep and requiring no local tooling installation?

khos2ow:
Those are 100% legitimate concerns @hanzei , and to be fair I cannot insist on this because I'm the author of that script anyways.

Using a `min_version` constraint is a real valid arguments but that was exactly the inception origin of `hugow`. At some point I was working on multiple websites with multiple different teams/people and neither of them were running the same Hugo version and I (and some other people involved) had to have multiple versions of Hugo (also _extended_ or normal) on my laptop to be able to run/test them locally which eventually led me to creating `hugow`.

@lieut-data and in the process I actually did consider using Docker container instead of writing my own script, but that's an extra dependency on the system, while `hugow` is _almost_ zero dependency script. And I figured for any non-technical person (which most probably be more involved in creating the content or maintaining the theme) using a single script was easier than using a docker container (not necessarily though TBH, but may be)

hanzei:
Mattermost is using Docker in the developers environment for the core product. Hence, I suppose a lot of developers have docker already installed.

khos2ow:
Fair enough, but as far as I know there's no official docker image for Hugo and it seems the most popular one is [klakegg/docker-hugo](https://github.com/klakegg/docker-hugo). Which is definitely less concerning (comparing to `hugow`) for the user of the tool, but at least it's self contained. I'm absolutely fine taking either approaches.

lieut-data:
@jaydeland / @metanerd, do you have an opinion on the best approach to take here? I'm personally leaning towards containerizing our build dependencies for the following reasons:
* we can pin the tooling, allowing us to easily build old versions that may otherwise require old hugo versions
* no manual installation of tools required (beyond Docker, which we already required for other contributions)
* easier to match the dev and CI builds to yield identical outputs

jaydeland:
@lieut-data and @khos2ow -> I agree that containerizing is the prefered approach. 


> @jaydeland / @metanerd, do you have an opinion on the best approach to take here? I'm personally leaning towards containerizing our build dependencies for the following reasons:
> 
>     * we can pin the tooling, allowing us to easily build old versions that may otherwise require old hugo versions
> 
>     * no manual installation of tools required (beyond Docker, which we already required for other contributions)
> 
>     * easier to match the dev and CI builds to yield identical outputs

lieut-data:
Thanks for spurring on this conversation @khos2ow! I'm going to close this pull request for now given the decision above.

khos2ow:
Yes sure, I'm gonna open another PR for docker based workflow based on [klakegg/docker-hugo](https://github.com/klakegg/docker-hugo) soon.

PR #517, opened 2020-01-24
Update link in "getting started" page

amyblais:
Fixes https://github.com/mattermost/mattermost-server/issues/13750.

I removed the signup link in case this bug happens again.


hanzei:
Thanks for the quick fix @amyblais :+1:

PR #387, opened 2019-10-10
Get rid of GOPATH in the developers setup guide

jorgeruvalcaba:
Removing GOPATH in the developer guide based on #380. Also, I changed the numeration in some guides.

Fixes #380

Issue #380, opened 2019-10-02
Get rid of GOPATH in the developers setup guide


hanzei:
Given we are now 100% using the local vendor folder with go modules, there is no need clone the repo inside `GOPATH` any longer. We should remove this from our setup guide.

Keep in mind that the server and webapp directories still have to be in the same directory so our `Makefile` can find them.

jorgeruvalcaba:
Can I work on this issue?

hanzei:
Sure, that would be awesome @jorgeruvalcaba  :tada:

jorgeruvalcaba:
Added a PR for this issue, thanks @hanzei

PR #519, opened 2020-01-26
Updated the webapp developer docs to match #387

mitchellroe:
#### Summary

In #387, we removed the instructions for placing your code in the GOPATH, but it looks like we missed the webapp instructions.  This just cleans those up to match.

#### Ticket Link

None.


mitchellroe:
Mentioning #380 for linking purposes.

PR #523, opened 2020-01-31
Add make run

hanzei:
#### Summary
- Place `.PHONY` in front of each target to ensure it doesn't get forgotten
- Make use of `hugo -s` to set the source
- Add `make run` as an alias of `make dev`, as `make run` is the more common term

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 .PHONY: dist 
-dist: #plugin-data
+dist: plugin-data
 	rm -rf ./dist
 	hugo -s site --destination ../dist/html
 
@@ -29,7 +29,7 @@ endif
 
 .PHONY: run
 run:
-	hugo server -s site
+	hugo server --buildDrafts -s site
 
 .PHONY: dev
 dev: run
--- a/Makefile
+++ b/Makefile
@@ -1,16 +1,18 @@
-.PHONY: dist plugin-data backend-plugin-data frontend-plugin-data
-
-dist: plugin-data
+.PHONY: dist 
+dist: #plugin-data
 	rm -rf ./dist
-	cd site && hugo --destination ../dist/html
+	hugo -s site --destination ../dist/html
 
+.PHONY: plugin-data
 plugin-data: backend-plugin-data frontend-plugin-data devtalks-data
 
+.PHONY: backend-plugin-data
 backend-plugin-data:
 	mkdir -p site/data
 	go run ./cmd/plugin-godocs > site/data/PluginGoDocs.json
 	go run ./cmd/plugin-manifest-docs > site/data/PluginManifestDocs.json
 
+.PHONY: frontend-plugin-data
 frontend-plugin-data:
 	rm -rf scripts/mattermost-webapp
 	cd scripts && git clone https://github.com/mattermost/mattermost-webapp.git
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -1,28 +1,35 @@
...
 	go run ./cmd/devtalks > site/data/DevTalks.json
 endif
 
-dev:
-	cd site && hugo server -D
+.PHONY: run
+run:
+	hugo server -s site
+
+.PHONY: dev
```
lieut-data: Maybe move this out into a `build/legacy.mk` so as to preserve compatibility but clearly deprecate?

Makefile
```diff
@@ -1,28 +1,35 @@
...
 	mkdir -p site/data
 ifdef YOUTUBE_API_KEY
 	go run ./cmd/devtalks > site/data/DevTalks.json
 endif
 
-dev:
-	cd site && hugo server -D
+.PHONY: run
+run:
+	hugo server -s site
```
lieut-data: Why no more `--buildDrafts`?

Makefile
```diff
@@ -1,28 +1,35 @@
-.PHONY: dist plugin-data backend-plugin-data frontend-plugin-data
-
-dist: plugin-data
+.PHONY: dist 
+dist: #plugin-data
 	rm -rf ./dist
-	cd site && hugo --destination ../dist/html
+	hugo -s site --destination ../dist/html
```
lieut-data: Should this now be `--destination dist/html`?

Makefile
```diff
@@ -1,28 +1,35 @@
-.PHONY: dist plugin-data backend-plugin-data frontend-plugin-data
-
-dist: plugin-data
+.PHONY: dist 
+dist: #plugin-data
```
lieut-data: No more `plugin-data`?

Makefile
hanzei: Good catch :+1:

Makefile
hanzei: To my own surprise: No. `destination` is relative to `-s`.

Makefile
hanzei: Fixed :+1:

Makefile
hanzei: How would moving it into a different file help with marking it as deprecated?

Makefile
hanzei: To be clear: I've tested the changes and they generate the files in the correct place.

Makefile
ali-farooq0: Any reason not to remove this rather than deprecating it?

Also the convention we've been using in other places is `run-server` should we do the same here?

ali-farooq0: though it should be `make run` but 🤷‍♂

hanzei: > Any reason not to remove this rather than deprecating it?

It might disrupt developers in there workflow if a command that they got used to using does not longer exist.

> Also the convention we've been using in other places is `run-server` should we do the same here?

Is this really the case? At last for MM server `make run` is the fire and forget command you want.

ali-farooq0: ahh true, i never use `run`, always use `run-server` for MM. Same with marketplace. Fair enough then `make run` it is.

Regarding `make dev`, at some point whenever we decide to remove `dev`, whoever was used to that workflow will get get disrupted. Either we remove it now or add some sort of warning message that this `make dev` is deprecated and encourage them to use `make run` instead, so atleast they'll know to switch.

hanzei: I would like to add a warning message but didn't found a good way of doing this. Do you have idea?

ali-farooq0: Not much of a `Makefile` expert but after some googling, found this to be working well.
```
dev:
	$(warning **** make dev is DEPRECATED, use make run instead ****)
	make run
```

hanzei: That technically works, but I spawns another process for `make run`. The purist in me doesn't like it, but I will go away sone anyway.

lieut-data: > How would moving it into a different file help with marking it as deprecated?

We do this in the server with [`build/legacy.mk`](https://github.com/mattermost/mattermost-server/blob/master/build/legacy.mk). I find it an elegant way to preserve backwards compatibility but avoid cluttering the primary `Makefile`.

PR #538, opened 2020-02-20
Update troubleshooting.md

J35u527:
Added an error message and a solution when running by first time make pre-run command and there is a missing gem called bundle


J35u527:
Requested changes goes to the following PR
https://github.com/mattermost/mattermost-developer-documentation/pull/539

PR #543, opened 2020-02-28
[MM-22700] Fix variadic functions on the plugin docs page

hanzei:
#### Summary
Fix variadic functions being shown as `[]`. I also did some cleanup of the code to keep it DRY.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22700

### DIFF
--- a/cmd/plugin-godocs/plugin-godocs.go
+++ b/cmd/plugin-godocs/plugin-godocs.go
@@ -101,7 +101,7 @@ func fields(list *ast.FieldList, info *types.Info) (fields []*Field) {
 			// If type is "...", t will start with [] instead of ...
 			// Replace it manually
 			if _, ok := x.Type.(*ast.Ellipsis); ok {
-				t = strings.ReplaceAll(t, "[]", "...")
+				t = strings.Replace(t, "[]", "...", 1)
 			}
 			field.Type = t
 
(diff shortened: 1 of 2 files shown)


cmd/plugin-godocs/plugin-godocs.go
```diff
@@ -85,29 +86,26 @@ func fields(list *ast.FieldList, info *types.Info) (fields []*Field) {
...
-					field.Type = "...interface{}"
-				}
-				fields = append(fields, field)
+			if xType == nil {
+				panic(fmt.Sprintf("type of %s is nil", field.Names))
 			}
+
+			t := xType.String()
+
+			// If type is "...", t will start with [] instead of ...
```
hanzei: I have no idea why `xType.String()` is e.g. `[]interface{}` for `LogError`. Hence, I created this workaround.

cmd/plugin-godocs/plugin-godocs.go
```diff
@@ -85,29 +86,26 @@ func fields(list *ast.FieldList, info *types.Info) (fields []*Field) {
...
+			// Replace it manually
+			if _, ok := x.Type.(*ast.Ellipsis); ok {
+				t = strings.ReplaceAll(t, "[]", "...")
```
lieut-data: If we were to accept a variadic number of slices, would this break? Should we just replace at the beginning?

hanzei: Good point. I forgot about variadic functions with slices. Changed the code to only replace once.

PR #590, opened 2020-04-07
Bump up minimum required Go version

agnivade:
#### Summary

Bump minimum required Go version to 1.14


grundleborg:
@agnivade in the developer setup guide we have various copy-paste commands that include a go version - these are out of date - please can you update or otherwise fix them?

PR #593, opened 2020-04-10
Update service.md

metanerd:
Documented service file did not work for me. The updated change works for me.


metanerd:
Seems like I used an outdated version and this should actually work.

PR #625, opened 2020-05-22
Replace esprima with espree

agarciamontoro:
#### Summary
@metanerd investigated the latest failed build in master and narrowed it down to this change: https://github.com/mattermost/mattermost-webapp/commit/397e4b65a610cfc81ab8076dbf9348c71d4d4872#diff-99e11b3b61425aff9ef2e561dd706b2cR600

After some more investigation, it looks like [esprima](https://github.com/jquery/esprima) does not recognize the [spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) in its latest release, even if they have it on master: https://github.com/jquery/esprima/issues/1927

This PR replaces esprima with [espree](https://github.com/eslint/espree), a more up-to-date fork.

See a bit more of context in this thread in Mattermost: https://community-daily.mattermost.com/core/pl/4tewph9mjpyfxywr9ygyg8syec

#### Ticket Link
--

### DIFF
--- a/scripts/plugin-jsdocs.js
+++ b/scripts/plugin-jsdocs.js
@@ -4,7 +4,7 @@ const espree = require('espree');
 
 // Parse the registry and extract the class methods, parameters and leading comments.
 const registryContent = fsp.readFileSync(path.join(__dirname, 'mattermost-webapp/plugins/registry.js'), 'utf-8')
-const registryParsed = espree.parse(registryContent, { comment: true, attachComment: true, sourceType: 'module', ecmaVersion: 10 });
+const registryParsed = espree.parse(registryContent, { comment: true, loc: true, sourceType: 'module', ecmaVersion: 10 });
 
 const pluginRegistryClassMethods = registryParsed.body.find(statement =>
     statement.type === "ExportDefaultDeclaration" &&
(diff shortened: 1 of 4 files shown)


scripts/plugin-jsdocs.js
```diff
@@ -1,10 +1,10 @@
 const path = require('path');
 const fsp = require('fs-promise');
-const esprima = require('esprima');
+const espree = require('espree');
 
 // Parse the registry and extract the class methods, parameters and leading comments.
 const registryContent = fsp.readFileSync(path.join(__dirname, 'mattermost-webapp/plugins/registry.js'), 'utf-8')
-const registryParsed = esprima.parseModule(registryContent, { comment: true, attachComment: true });
+const registryParsed = espree.parse(registryContent, { comment: true, attachComment: true, sourceType: 'module', ecmaVersion: 10 });
```
agarciamontoro, line 7: I set the `ecmaVersion` as `10` because it works (it also works with 9, but not with 8 or older). But I don't really know what we should set it to.

metanerd:
> This seems excellent, @agarciamontoro! Beyond fixing the build, are they any observable differences in the final output (i.e. before the most recent webapp changes)?

Where is this visible in the normal docs, @lieut-data ?

lieut-data:
@metanerd, this would mostly be https://developers.mattermost.com/extend/plugins/webapp/reference/.

agarciamontoro:
That's a good question for which I don't have an answer yet. I'll check and come back with some more information!

metanerd:
comments_missing
The comments seem to be unfortunately missing.

agarciamontoro:
I could reproduce that, too. And looks like espree handles comments differently. From its README:

> The primary goal is to produce the exact same AST structure and tokens as Esprima, and that takes precedence over anything else. (The AST structure being the ESTree API with JSX extensions.) Separate from that, Espree may deviate from what Esprima outputs in terms of where and how comments are attached, as well as what additional information is available on AST nodes.

agarciamontoro:
They're back: 
image

espree does not attach the comments to its adjacent methods, so I needed to do that manually.

site/layouts/shortcodes/pluginjsdocs.html
```diff
@@ -29,7 +29,7 @@ <h5 id="{{ .Name }}">{{ .Name }}</h3>
 {{ define "JSFunctionSignatureHTML" }}
     <pre><code>/**
 {{- range .Comments }}
-    * {{ . }}
+    *{{ . }}
```
agarciamontoro, line 32: I've removed the `trim` call in the comments strings, so there were two spaces here. I think it is better to rely on the formatting of the comments, as they sometimes use it in order to make the text clearer. See, for example:
image

agarciamontoro:
I've noticed that the optional parameters are not rendered, as their object structure in the AST is different from regular parameters. I've opened a different issue for this: https://github.com/mattermost/mattermost-developer-documentation/issues/627

@hanzei, you are the fastest reviewer and caught me in the middle of writing this comment :joy: That bug seems to be old, see the above issue.

agnivade:
This was much appreciated @agarciamontoro ! Thanks.

PR #662, opened 2020-07-23
Add GH Action to deploy the hugo site to gh-pages as a subdir for each contributor

jaydeland:
This change will update a dedicated subdirector within gh-pages for the github authour that triggered the PR with the content of the documentation site. This will give each contributor a dedicated area to preview the changes in the PR without the use of Vercel. Our legal team has concerns with the Vercel terms of service.
@justinegeffen - will this meet your needs?


justinegeffen:
Thanks, @jaydeland! I understand Legal's concerns and am on board with that. This sounds like it would work well - would we be able to use it on docs.mattermost.com, which uses Sphinx? This is the conf.py file: https://github.com/mattermost/docs/blob/master/source/conf.py.

.github/workflows/gh-pages.yml
```diff
@@ -0,0 +1,40 @@
...
+          publish_dir: ./dist/html
+          publish_branch: gh-pages
+          force_orphan: true
+      - name: Add comment to PR
+        uses: peter-evans/create-or-update-comment@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          issue-number: ${{ github.event.number }}
+          body: |
+            Published to [preview environment](https://mattermost.github.io/mattermost-developer-documentation/${{ github.actor }})
```
metanerd, line 39: Deployment looks great! I guess this path would need to be more generic.

jaydeland:
Yes - the Action used to create the site has a setting for Sphinx that I can setup if this works.

> Thanks, @jaydeland! I understand Legal's concerns and am on board with that. This sounds like it would work well - would we be able to use it on docs.mattermost.com, which uses Sphinx? This is the conf.py file: https://github.com/mattermost/docs/blob/master/source/conf.py.

.github/workflows/gh-pages.yml
jaydeland: I was thinking that one path per contributor was enough, but we could use PR # if we clean them up after its closed.

justinegeffen:
> Yes - the Action used to create the site has a setting for Sphinx that I can setup if this works.
> 
> > Thanks, @jaydeland! I understand Legal's concerns and am on board with that. This sounds like it would work well - would we be able to use it on docs.mattermost.com, which uses Sphinx? This is the conf.py file: https://github.com/mattermost/docs/blob/master/source/conf.py.

Thank you! I've tested the preview out today and it works really well (and eliminates a few issues I had with the previous solution). 👍

PR #665, opened 2020-07-29
Adding we’re hiring notification banner

asaadmahmood:
#### Summary
Adding we’re hiring notification banner

Screenshot 2020-07-29 at 1 15 01 PM


site/static/css/bar.css
```diff
@@ -1,15 +1,44 @@
...
+    right: 0px;
+    opacity: .3;
+    top: 0;
+    width: 64px;
+    line-height: 64px;
+}
+
+.notification-bar-hide:hover {
+    opacity: .5;
+}
```
hanzei, line 44: Nit: Should there be an empty line?

asaadmahmood: Yup, that's what I was thinking about as well, but I guess that's the way it is.

PR #690, opened 2020-09-17
Correcting info on starting the webserver for the developer version

aspleenic:
Correcting an issue in the developer setup instructions. The original instruction said to run `make run` in the mattermost-webapp directory, but it needs to be run in mattermost-server


sudheerDev:
@aspleenic Hey, `make run` from webapp is approrpriate here. Though I see there is confusion here as server needs to be running with `make run-server` inorder to see UI with `make run` in webapp but the changes are not appropriate as the usual workflow most devs follow is to run server and client independent of eachother.

PR #700, opened 2020-10-03
Update gh-pages.yml

jaydeland:
Add check to see if the actor for the PR is a Mattermost Org member


jaydeland:
This PR adds a check to see if the user/actor is a member of the GitHub Organization and if they are not it will not deploy the github pages preview site. It also removes the s3 Publishing in CircleCI as this is no longer needed.

PR #711, opened 2020-10-14
Detox/E2E: Add more details about TM4J

saturninoabril:
#### Summary
Screen Shot 2020-10-14 at 9 44 33 AM


enahum (approved):
Looks good.

I wonder if there is a better way to do this? Maybe use the same ID as the original ticket? For example MM-251526

Why? Because tickets and help wanted tickets are not linked to the test tickets and it can be a little bit cumbersome

saturninoabril:
For me, two reasons for using TM4J:
1. Easy to integrate test results to Test Management since it directly translates/communicates to TM4J.  In the past with Cypress, we used to have Jira ticket ID as test key but there's no easy way to integrate with TM4J.
2. Easy to map manually when validating/updating test cases.

Per my observation, once the test is automated, we are no longer looking at which Jira issue / HW it was resolved. Test maintenance involves direct mapping between TM4J and automated script.

josephbaylon (approved):
LGTM

justinegeffen (approved):

justinegeffen:
Apologies for the delay! LGTM, thanks!

saturninoabril:
Thanks @justinegeffen and no worries.

PR #727, opened 2020-10-31
Update footer button transitions to match other website elements

maticbasle:
This PR updates the footer buttons to include transitions to match other website elements such as the buttons on the index page.

### DIFF
--- a/site/layouts/partials/footer.html
+++ b/site/layouts/partials/footer.html
@@ -2,9 +2,9 @@
     <div class="container-fluid">
         <div class="footer__copyright">© Mattermost, Inc. All Rights Reserved.</div>
         <div class="footer__icons">
-            <a href="{{ .Site.Params.footer.facebook }}" target="_blank" rel="noopener noreferrer" class="btn-primary" style="transition: background-color .15s ease-in-out"><i class="fa fa-facebook"></i></a>
-            <a href="{{ .Site.Params.footer.twitter }}" target="_blank" rel="noopener noreferrer" class="btn-primary" style="transition: background-color .15s ease-in-out"><i class="fa fa-twitter"></i></a>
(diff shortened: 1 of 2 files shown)


site/layouts/partials/footer.html
```diff
@@ -2,9 +2,9 @@
     <div class="container-fluid">
         <div class="footer__copyright">© Mattermost, Inc. All Rights Reserved.</div>
         <div class="footer__icons">
-            <a href="{{ .Site.Params.footer.facebook }}" target="_blank" rel="noopener noreferrer"><i class="fa fa-facebook"></i></a>
-            <a href="{{ .Site.Params.footer.twitter }}" target="_blank" rel="noopener noreferrer"><i class="fa fa-twitter"></i></a>
-            <a href="{{ .Site.Params.footer.youtube }}" target="_blank" rel="noopener noreferrer"><i class="fa fa-youtube"></i></a>
+            <a href="{{ .Site.Params.footer.facebook }}" target="_blank" rel="noopener noreferrer" class="btn-primary" style="transition: background-color .15s ease-in-out"><i class="fa fa-facebook"></i></a>
```
hanzei, line 5: Good catch :+1: 

What do you think about moving the `transition: background-color .15s ease-in-out` into `style.css`?

hanzei (changes requested):
Thanks for the PR :+1: 

One suggestion on the code organisation

site/layouts/partials/footer.html
maticbasle: To apply the transition to all primary buttons or create a new class?

hanzei: Can we add it to the `footer__icons` class?

hanzei (approved):
LGTM :+1:

abhijit-singh (approved):

PR #730, opened 2020-11-08
SEO Updates for canonical URLs, robots.txt, moved app scripts to footer

lawrencejohnson:
#### Summary
This pull request includes configuration changes and modifications to the global header and footer files based on requests for SEO optimizations. The configuration file for the site has been modified to use absolute URLs for links, sitemap, canonical meta tag, and app assets like css and javascript files.

- config.toml updated to use absolute URLs, affects page links, meta tags, script and css references, and sitemap.xml
- head.html revised to better leverage absolute url paths, scripts moved to footer.html
- footer.html app javascript references moved from head.html to reduce render blocking scripts on page load
- robots.txt added per specification provided by SEO team

#### JIRA Tickets
https://mattermost.atlassian.net/browse/MAR-158
https://mattermost.atlassian.net/browse/MAR-163
https://mattermost.atlassian.net/browse/MAR-174
https://mattermost.atlassian.net/browse/MAR-179

### DIFF
--- a/site/config.toml
+++ b/site/config.toml
@@ -1,6 +1,8 @@
 # Page settings
-baseURL = ""
-relativeURLs = true
+baseURL = "https://developers.mattermost.com/"
+canonifyURLs = true
+#relativeURLs = true
+
 title = "Mattermost Developers"
 languageCode = "en-us"
 publishDir = "../docs"
--- a/site/layouts/partials/footer.html
+++ b/site/layouts/partials/footer.html
@@ -8,3 +8,11 @@
         </div>
     </div>
 </footer>
+
+<!-- JS -->
+<script
+  src="https://code.jquery.com/jquery-3.5.1.min.js"
+  integrity="sha256-9/aliU8dGd2tb6OSsuzixeV4y/faTqgFtohetphbbj0="
+  crossorigin="anonymous"></script>
+<script src="{{ "js/tabs.js" | absURL }}"></script>
+<script src="{{ "js/main.js" | absURL }}"></script>
--- a/site/layouts/partials/head.html
+++ b/site/layouts/partials/head.html
@@ -13,22 +13,18 @@
 <link rel="canonical" href="{{ .Permalink }}">
 {{ end }}
 
-<link rel="shortcut icon" type="image/png" href="{{ .Site.BaseURL }}/img/favicon-32x32.png" />
+<link rel="shortcut icon" type="image/png" href="{{ "img/favicon-32x32.png" | absURL }}" />
 
 <link href="https://fonts.googleapis.com/css?family=Open+Sans:300,400,600" rel="stylesheet">
 
 <!-- CSS  -->
-<link rel="stylesheet" href="{{ .Site.BaseURL }}/css/bootstrap.min.css">
-<link rel="stylesheet" href="{{ .Site.BaseURL }}/css/tabs.css">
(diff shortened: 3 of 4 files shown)


site/config.toml
```diff
@@ -1,6 +1,8 @@
 # Page settings
-baseURL = ""
-relativeURLs = true
+baseURL = "https://developers.mattermost.com/"
```
lawrencejohnson, line 2: Modification of baseURL was tested using hugo server option `--baseURL=https://developers.mattermost.com`

site/layouts/partials/footer.html
```diff
@@ -8,3 +8,11 @@
         </div>
     </div>
 </footer>
+
+<!-- JS -->
+<script
```
lawrencejohnson, line 13: jQuery updated to latest version of `3.x`

site/layouts/partials/head.html
```diff
@@ -13,22 +13,18 @@
...
+<link rel="stylesheet" href="{{ "css/tabs.css" | absURL }}">
+<link rel="stylesheet" href="{{ "css/bar.css" | absURL }}">
+<link rel="stylesheet" href="{{ "css/styles.css" | absURL }}">
+<link rel="stylesheet" href="{{ "css/code.css" | absURL }}">
+<link rel="stylesheet" href="{{ "css/note.css" | absURL }}">
 
-<!-- JS -->
-<script src="https://ajax.googleapis.com/ajax/libs/jquery/3.2.1/jquery.min.js"></script>
-<script src="{{ .Site.BaseURL }}/js/tabs.js"></script>
-<script src="{{ .Site.BaseURL }}/js/main.js"></script>
```
lawrencejohnson, line 31: Scripts moved to `footer.html`

site/config.toml
```diff
@@ -1,6 +1,8 @@
...
+canonifyURLs = true
+#relativeURLs = true
```
hanzei, line 4: Can we just remove this?

site/static/robots.txt
```diff
@@ -0,0 +1,8 @@
+User-Agent: *
+Disallow: /cgi-bin/
+Disallow: /&utm_source
+Disallow: /&utm_medium
+Disallow: /&utm_campaign
+Disallow: /?utm_
+
+Sitemap: https://developers.mattermost.com/sitemap.xml
```
hanzei, line 8: Non blocking idea: Does it make sense to generate this file based on the configured SiteURL instead of having a static one? This is easily possible with Hugo: https://gohugo.io/templates/robots/

hanzei (approved):
Great work. Thanks for updating the website :+1: 

Two non blocking comments below

site/config.toml
lawrencejohnson: If you're referencing the commented `#relativeURLs` option, yes that's fine.

site/static/robots.txt
lawrencejohnson: My opinion is that it should not be auto-generated since the SEO team can generate these based on the specific needs they are looking for. Having the configured SiteURL pull into the Sitemap reference only adds the benefit of if the site is moved to a different URL, but comes at the cost of the robots.txt file needing to be audited for changes that an auto-generated system might make without warning (should the robots engine change in a future update).

jasonblais (approved):

PR #741, opened 2020-12-04
Add blog post about common i18n mistakes

hmhealey:
I've wanted to write this for a while, and since we hit one of these mistakes recently, I thought I'd write it down finally so that I can just link to it instead of repeating myself periodically.

I've got some other ideas for i18n-related blog posts in the future as well such as how to deal with FormattedMessage vs injectIntl vs useIntl vs localizeMessage and that pseudolocalization thing I hacked together way back for finding these sorts of mistakes.


site/content/blog/2020-12-04-common-i18n-mistakes.md
```diff
@@ -0,0 +1,302 @@
...
+                </a>
+            ),
+        }}
+    />
+);
+```
+
+You'll often see this used in older Mattermost code to solve this problem. It allows us to add more complex formatting within translated text without making it impossible to translate, but it does make it harder for translators to understand the meaning of the text. They need to spend time reconstructing the entire string which may be difficult. It also makes the code more complicated to read since developers have to worry about more React elements.
+
+### Solution 2: FormattedHTMLMessage
```
calebroseland, line 214: @hmhealey I think this component/API was dropped altogether in v5

site/content/blog/2020-12-04-common-i18n-mistakes.md
```diff
@@ -0,0 +1,302 @@
...
+
+Note that the React example specifies that `remaining` is a number since otherwise zero would be treated as an empty string because it's falsey.
+
+As mentioned before, these are both technically correct since they add the number of remaining emails to the right place, but they both look ugly. Instead, both libraries support pluralization by passing in a count of some object which is the number of remaining emails in this case.
+
+```jsx
+return (
+    <FormattedMessage
+        id='email_sender.remaining'
+        defaultMessage='{remaining, number} {remaining, plural, one {email} other {emails}} remaining.'
```
enahum, line 124: not sure if you want to add this as an example in Spanish there is a plural word for `remaining`.
singular: restante
plural: restantes

Although you can translate it to something like
`{remaining, number} {remaining, plural, one {correo pendiente} other {correos pendientes}}.`

site/content/blog/2020-12-04-common-i18n-mistakes.md
enahum, line 214: Maybe avoid this one as it only applies to web?

enahum (changes requested):
great post, I just would avoid the use of `FormattedHTMLMessage`, also as @calebroseland said, I think this one was removed from react-intl but haven't confirmed.

site/content/blog/2020-12-04-common-i18n-mistakes.md
hmhealey: Good idea. That'll also help reinforce how other languages do things differently. I considered trying to add some non-English examples myself, but my French is really rusty.

site/content/blog/2020-12-04-common-i18n-mistakes.md
```diff
@@ -0,0 +1,315 @@
...
+categories:
+    - "i18n"
+author: Harrison Healey
+github: hmhealey
+community: harrison
+---
+
+Languages are complicated, and every language is complicated in different ways that can be hard to understand without learning every single one of them. Some languages form words from multiple characters while others have symbols that represent entire concepts. Some feature words without pluralization or gender and rely on context for that while others have two or even more genders for words. Some are very phonetic while others pronounce words seemingly at random (*cough* English, though *cough*).
+
+Thanks to a tremendous contribution long ago from then community member Elias Nahum, we have full support for translation throughout Mattermost, and thanks to our community of translators, Mattermost is usable by many people speaking a variety of different languages.
```
justinegeffen, line 14:
```suggestion
Thanks to a tremendous contribution long ago from then-community member Elias Nahum, we have full support for translation throughout Mattermost, and thanks to our community of translators, Mattermost is usable by many people speaking a variety of different languages.
```

site/content/blog/2020-12-04-common-i18n-mistakes.md
```diff
@@ -0,0 +1,315 @@
...
+
+Because we support so many languages, we have to be aware of how to keep our applications properly translatable. There's many easy mistakes you can make when writing an application in one language that can make it difficult to translate into others. We use [react-intl](https://formatjs.io/docs/react-intl/) in our client-side applications and [go-i18n](https://github.com/nicksnyder/go-i18n) on the server which both offer a variety of features to help with this, but we can still cause problems if we're not careful when using translated text. Here's a few examples of common problems that we've run into and how to solve them.
```
justinegeffen, line 16:
```suggestion
Because we support so many languages, we have to be aware of how to keep our applications properly translatable. There are many easy mistakes you can make when writing an application in one language that can make it difficult to translate into others. We use [react-intl](https://formatjs.io/docs/react-intl/) in our client-side applications and [go-i18n](https://github.com/nicksnyder/go-i18n) on the server which both offer a range of features to help with this, but we can still cause problems if we're not careful when using translated text. Here are a few examples of common problems that we've run into and how to solve them.
```

site/content/blog/2020-12-04-common-i18n-mistakes.md
```diff
@@ -0,0 +1,315 @@
...
+
+## Mistake 1: Not translating something
+
+This one is definitely the easiest to avoid, but it can still catch people who aren't used to writing applications outside their native language.
+
+For example, suppose someone is adding a new button to the web app using React. If they forget to make it to use react-intl, they may write something like this:
```
justinegeffen, line 22:
```suggestion
For example, suppose someone is adding a new button to the web app using React. If they forget to make it to use ``react-intl``, they may write something like this:
```

justinegeffen: Not sure if the formatting suggestion works in this instance, so feel free to ignore if not! I haven't suggested it going forward for that reason.

site/content/blog/2020-12-04-common-i18n-mistakes.md
```diff
@@ -0,0 +1,315 @@
...
+        </button>
+    );
+}
+```
+
+The code ends up being a bit longer, but that's more than worth it for making the app translatable.
+
+## Mistake 2: Hard coding translations
+
+This mistake is a bit rarer since it requires mixing translated and non-translated text, but it comes from a real-world example in Mattermost.
```
justinegeffen, line 51:
```suggestion
This mistake is less common since it requires mixing translated and non-translated text, but it comes from a real-world example in Mattermost.
```

site/content/blog/2020-12-04-common-i18n-mistakes.md
```diff
@@ -0,0 +1,315 @@
...
+);
+```
+
+You'll often see this used in older Mattermost code when we're trying to format part of a string. It allows us to add more complex formatting within translated text without making it impossible to translate.
+
+It does, however, make it harder for translators to understand the meaning of the text. They need to spend time reconstructing the entire sentence which may be difficult. It also makes the code more complicated to read for the same reason.
+
+### Solution 2: FormattedHTMLMessage
+
+`FormattedHTMLMessage` was another way to solve this problem, but it was removed in more recent versions. It's a compoent provided by react-intl, similar to `FormattedMessage`, but it allows us to include HTML in the translation string instead of nesting translated text. Since it's been removed, it can't be used any more, but it demonstrates another problem with ran into in the past.
```
calebroseland, line 226:
```suggestion
`FormattedHTMLMessage` was another way to solve this problem, but it was removed in more recent versions. It's a component provided by react-intl, similar to `FormattedMessage`, but it allows us to include HTML in the translation string instead of nesting translated text. Since it's been removed, it can't be used any more, but it demonstrates another problem with ran into in the past.
```

site/content/blog/2020-12-04-common-i18n-mistakes.md
hmhealey: I'll actually rewrite references to that as React Intl since that's how they write it in their docs. For go-i18n, they just refer to it as that, so I don't think it needs to use preformatted text.

Also, I didn't realize you could do inline code with more than one backquote despite all my time working with it 😛

site/content/blog/2020-12-04-common-i18n-mistakes.md
```diff
@@ -0,0 +1,315 @@
...
+return (
+    <FormattedMessage
+        id='post.commentedOn'
+        defaultMessage='Commented on {username} message:'
+        values={{username}}
+    />
+)
+```
+
+Here, we're using react-intl to add Billy's name into the text as a value, but we're making a mistake here by trying to make the username into a possessive adjective ourselves. By manually adding the "'s", we're applying English grammar regardless of the user's language setting, so this label won't make sense in other languages.
```
justinegeffen, line 70:
```suggestion
Here, we're using react-intl to add Billy's name into the text as a value, but we're making a mistake by trying to make the username into a possessive adjective ourselves. By manually adding the "'s", we're applying English grammar regardless of the user's language setting, so this label won't make sense in other languages.
```

site/content/blog/2020-12-04-common-i18n-mistakes.md
```diff
@@ -0,0 +1,315 @@
...
+
+Instead, the "'s" should be part of the translation string. That required changing our grammar rules slightly since we previously avoided adding "'s" to the end of names that ended in the letter S, but the new way is still correct by other popular grammar guidelines. Isn't English great?
```
justinegeffen, line 72:
```suggestion
Instead, the "'s" should be part of the translation string. That required changing our grammar rules slightly since we previously avoided adding "'s" to the end of names that ended in the letter "s", but the new way is still correct according to other popular grammar guidelines. Isn't English great?
```

site/content/blog/2020-12-04-common-i18n-mistakes.md
```diff
@@ -0,0 +1,315 @@
...
+*/
+
+func logEmailsRemaining(remaining int) {
+    log.Print(translateFunc("app.email_sender.remaining", map[string]interface{}{
+        "Remaining": remaining,
+    }))
+}
+```
+
+Note that the React example uses a feature of react-intl to specify that `remaining` is a number. Without it, "0 emails remaining" wouldn't be shown properly because it's a falsey value in JavaScript.
```
justinegeffen, line 116:
```suggestion
Note that the React example uses a feature of react-intl to specify that `remaining` is a number. Without it, "0 emails remaining" wouldn't be shown properly because it's a false value in JavaScript.
```

site/content/blog/2020-12-04-common-i18n-mistakes.md
```diff
@@ -0,0 +1,315 @@
...
+
+That is an example of how we can tell react-intl to behave slightly differently though. If we use the `plural` modifier, we can specify different text based on the value of `remaining`. go-i18n has a similar feature, but it takes two different versions of the translated string instead of constructing a one that is more complicated.
```
justinegeffen, line 118:
```suggestion
That is an example of how we can tell react-intl to behave slightly differently though. If we use the `plural` modifier, we can specify different text based on the value of `remaining`. go-i18n has a similar feature, but it takes two different versions of the translated string instead of constructing one that's more complicated.
```

site/content/blog/2020-12-04-common-i18n-mistakes.md
justinegeffen, line 226:
```suggestion
`FormattedHTMLMessage` was another way to solve this problem, but it was removed in more recent versions. It's a component provided by react-intl, similar to `FormattedMessage`, but allows us to include HTML in the translation string instead of nesting translated text. Since it's been removed, it can't be used any more, but it demonstrates another problem we ran into in the past.
```

site/content/blog/2020-12-04-common-i18n-mistakes.md
```diff
@@ -0,0 +1,315 @@
...
+    />
+);
+```
+
+As with the previous solution, this allows translators to reconstruct the sentence however they need, but it comes with a few serious downsides such as:
+1. Translators needed to understand HTML, including more complicated things like the extra parameters on the `a` tag.
+2. It added the chance that an incorrect translation could break part of the application by accidentally including malformed HTML within a string.
+3. It also had potential performance impacts since we were going outside of React's DOM manipulation optimizations by using raw HTML.
+4. We lost the ability to wrap parts of the string in React elements which can have more functionality, such as encapsulating the additional parameters passed into `a` tag.
+5. Since it used HTML, `FormattedHTMLMessage` wasn't available in `react-native` for use in our mobile apps.
```
justinegeffen, line 244:
```suggestion
5. Since it used HTML, `FormattedHTMLMessage` wasn't available in react-native for use in our mobile apps.
```

justinegeffen: Not sure my correction of react-native is correct - based on other instances of react mentions without formatting.

site/content/blog/2020-12-04-common-i18n-mistakes.md
```diff
@@ -0,0 +1,315 @@
...
+
+But despite these issues, it helped point us in the right direction since it lead to a custom component that is much nicer to work with.
```
justinegeffen, line 246:
```suggestion
But despite these issues, it helped point us in the right direction since it led to a custom component that's much nicer to work with.
```

site/content/blog/2020-12-04-common-i18n-mistakes.md
```diff
@@ -0,0 +1,315 @@
...
+        }}
+    />
+);
+```
+
+## Closing Thoughts
+
+Hopefully, we can do our best to make Mattermost easily translatable and available in many different languages, even as Mattermost grows and the amount of text that needs to be translated increases. By avoiding these mistakes and using some of these techniques, we can help to reduce the workload of our translators and make their jobs easier.
+
+Special thanks again to everyone who help translate Mattermost and who make it usable by so many more people around the world. We tremendously appreciate your efforts and all the work you put in to make sure the translations stay complete and up to date.
```
justinegeffen, line 313:
```suggestion
Special thanks again to everyone who helps translate Mattermost and who make it usable by so many more people around the world. We tremendously appreciate your efforts and all the work you put in to make sure the translations stay complete and up to date.
```

justinegeffen (approved):

justinegeffen:
Great blog, @hmhealey! A few suggestions and minor corrections but otherwise LGTM. :)

site/content/blog/2020-12-04-common-i18n-mistakes.md
hmhealey: I'll also change this to React Native since that's the proper name

site/content/blog/2020-12-04-common-i18n-mistakes.md
hmhealey: Falsy is actually a correct word here, although I just learned it's usually "falsy" without the e. It's a JavaScript term for things that are considered to be false like the number 0.

enahum (approved):

jaydeland (approved):

PR #758, opened 2021-01-22
Modified `smali` related paths based apktool change

sadohert:
apktool seems to unpack the apk files to different paths.  Admittedly I don't understand the inner workings of whats happening, but I noticed the unpacked apk had smali files at a different path than what was expected by the `sign-android` script.

My `apktool` version (in case its relevant):
```
Apktool v2.5.0 - a tool for reengineering Android apk files
with smali v2.4.0 and baksmali v2.4.0
```
I also shifted `mv $SRC $DEST` down a few lines as this seemed to be done too early for the `sed` modifications to apply.

### DIFF
--- a/site/static/scripts/sign-android
+++ b/site/static/scripts/sign-android
@@ -58,7 +58,7 @@ fi
 
 ANDROID="${ANDROID_HOME:-$ANDROID_SDK_ROOT}"
 
-if [ -z $ANDROID ]; then 
+if [ -z $ANDROID ]; then
     echo ""
     echo "\$ANDROID_HOME or \$ANDROID_SDK_ROOT environment variable not set."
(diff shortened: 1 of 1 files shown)


sadohert:
Hi @enahum - I tested the original version of the script with Apktool v2.4.1 and I wasn't able to make it work for Android.

I tested the changes I'm proposing in this PR with both v2.4.1 and v2.5.0 and was able to successfully install and run Mattermost.

enahum (approved):

sadohert:
@enahum  anyone else from engineering need to review?  Or could @justinegeffen / @cwarnermm be the 2nd reviewer?

enahum:
This is good to merge, no need for a second review of any kind

PR #778, opened 2021-03-19
Update contribution-checklist.md

amyblais:
I've heard feedback from QA that devs are not always adding QA steps or expected results when they submit a PR, and then QA needs to go after the devs to find out the test steps. Having the test steps ready saves dev and QA time.

I'm updating the docs to add reminders for devs to add QA test steps to PRs (if the test steps are not already listed in the Jira ticket). Additionally we could also add a note about this to the PR Templates in the different repos.

In the past we tried to make it so that test steps would be required to be added when moving a Jira ticket to "PR Submitted", but it caused some issue with the Jira flow https://community.mattermost.com/core/pl/dnh9ibw56fypzbj9n3imybo41h. Also, not all tickets/PRs need QA testing.

### DIFF
--- a/site/content/contribute/getting-started/contribution-checklist.md
+++ b/site/content/contribute/getting-started/contribution-checklist.md
@@ -28,6 +28,7 @@ Follow this checklist for submitting a pull request (PR):
 6. The PR title begins with the Jira or GitHub Ticket ID (e.g. `[MM-394]` or `[GH-394]`) and summary template is filled out.
 7. If your PR adds or changes a RESTful API endpoint, please update the [API documentation](https://github.com/mattermost/mattermost-api-reference).
 8. If your PR adds a new plugin API method or hook, please add an example to the [Plugin Starter Template](https://github.com/mattermost/mattermost-plugin-starter-template).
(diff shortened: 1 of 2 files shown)


site/content/contribute/getting-started/contribution-checklist.md
```diff
@@ -28,6 +28,7 @@ Follow this checklist for submitting a pull request (PR):
 6. The PR title begins with the Jira or GitHub Ticket ID (e.g. `[MM-394]` or `[GH-394]`) and summary template is filled out.
 7. If your PR adds or changes a RESTful API endpoint, please update the [API documentation](https://github.com/mattermost/mattermost-api-reference).
 8. If your PR adds a new plugin API method or hook, please add an example to the [Plugin Starter Template](https://github.com/mattermost/mattermost-plugin-starter-template).
+8. If QA review is applicable, your PR includes test steps or expected results.
```
lieut-data, line 31: nit: can we also bump the numbers here, for the "plain text readers"?

lieut-data (approved):

sbishel (approved):

PR #789, opened 2021-04-13
typo for osslsigncode

benarent:
#### Summary
I've been trying to use `osslsigncode` and used your instructions. I found a few typos which are fixed here.


cwarnermm (approved):
From a technical perspective, I can't add much value, but approving!

Willyfrog (approved):
thanks for taking the time to upgrade the docs, you are right that this is the right way to sign the executable.

hanzei:
Thanks @benarent :+1:

PR #793, opened 2021-04-19
Add info about enabling plugin during manual installation

komik966:

### DIFF
--- a/site/content/extend/plugins/webapp/hello-world.md
+++ b/site/content/extend/plugins/webapp/hello-world.md
@@ -174,6 +174,15 @@ Install the plugin in one of the following ways:
                 plugin.json
                 main.js
     ```
+    - Enable the plugin in ``config.json``:
+      
+    ```
+    "PluginStates": {
+        "com.mattermost.webapp-hello-world": {
+          "Enable": true
+        }
+    },
+    ```
     - Restart the Mattermost server.
 
 Navigate to a regular Mattermost page and observe the new icon in the channel header. Click the icon and observe the alert dialog.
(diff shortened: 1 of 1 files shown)


mickmister:
Thanks for the contribution @komik966!

However I don't think the step added in this PR is necessary to enable/disable a plugin. You should be able to click `Enable` or `Disable` in the system console to enable/disable individual plugins. Can you try this to make sure this works in your environment?

image

komik966:
@mickmister Yes, this also works in my environment.
When I followed current documentation about manual plugin installation - it didn't worked. After debugging I realized that plugin can be enabled / disabled :)
So, I would consider adding this step but describe here that plugin needs to be enabled in system console (not through `config.json` as I described)

hanzei:
@mickmister The documents mentions the starter template: 

> Note that the steps below are intentionally very manual to explain all of the pieces fitting together. In practice, we recommend referencing mattermost-plugin-starter-template for helpful build scripts.

2/5 that this phrase is enough and we should keep the document as it is.

mickmister (approved):
LGTM :+1: Thanks @komik966!

hanzei (approved):

site/content/extend/plugins/webapp/hello-world.md
```diff
@@ -175,5 +175,6 @@ Install the plugin in one of the following ways:
                 main.js
     ```
     - Restart the Mattermost server.
+    - Enable the plugin in System Console UI
```
justinegeffen, line 178:
```suggestion
    - Enable the plugin in **System Console > Plugins (Beta) > Plugin Management**.
```

justinegeffen:
Thanks for this, @komik966! I've provided a suggestion for the workflow. LMK if you have any questions!

komik966:
@justinegeffen Good suggestion. I've updated.

justinegeffen (approved):

PR #800, opened 2021-04-23
Fix LHS menu

hanzei:
#### Summary
Fix unfolding of the LHS menu

#### Ticket Link
None

PR #373, opened 2019-09-30
Sidebar improvements

lieut-data:
The developer documentation has been broken locally for me for a while, though curiously not broken on mattermost.dev:
image

Fix this, along with an issue that did impact mattermost.dev re: "Best Practices" being misnamed. This is part of some initial tidying before we begin to spend some mana improving the actual contents here.

You may find the individual commits easier to review re: whitespace.

PR #801, opened 2021-04-24
[MM-35160] Use file structure instead of Params for LHS menu

hanzei:
#### Summary
Instead of relying on the `Section` and `Subsection` params which are error prone the LHS now uses the file structure to render the LHS menu.

While touching the code I did some cleanup like removing `.Scratch` and simplifying the code.

I tried to reduce the code duplication, but using a partial is tricky, as there is a lot of context to pass to it. Given that all code is in one file I think it's fine to have the duplication. 

The real changes are in https://github.com/mattermost/mattermost-developer-documentation/commit/b0e5b9487dc0eda361e14e1497d627bb9d7b6b77.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35160


hanzei:
@lieut-data Requested a review from you as you did the work on https://github.com/mattermost/mattermost-developer-documentation/pull/373 and I don't know about anyone else knowable about Hugo.

lieut-data:
For context:

> Instead of relying on the Section and Subsection params which are error prone the LHS now uses the file structure to render the LHS menu.

Why are these error prone?

hanzei:
@lieut-data It's not obvious for a writer if they got the menu params wrong. E.g. https://developers.mattermost.com/extend/plugins/server/debugging/ is missing them and hence the LHS menue doesn't correctly open

Screenshot from 2021-04-27 09-11-21

I was burned by the same issue recently (fixed in https://github.com/mattermost/mattermost-developer-documentation/pull/800) when I confused ´subsection` with `section`.


The PR also fixed some edge cases where multiple page with the same title would all get highlighted:
Screenshot from 2021-04-27 09-12-46

lieut-data (approved):
Sweet!

hanzei:
@justinegeffen Could you please do a QA review for the PR and double check that the LHS menu renders correctly on all pages?

justinegeffen:
It looks good to me, thanks Ben!

justinegeffen (approved):

PR #846, opened 2021-06-04
Nvm npm 7

neallred:
#### Summary
Node 14 is bundled with npm 6. npm 6 only knows about lockfile version 1 in package-lock.json, and so it ends up making a lot of changes to the lock file that the user didn't intend when doing an `npm install`. When you install & manage node/npm with `nvm`, you can  nvm install-latest-npm to manually install a version >= 7, which understands & does not modify the newer lock file.

#### Ticket Link

No known ticket


hmhealey:
Thanks for the PR, @neallred! I didn't realize I'd forgotten to update these docs.

We actually went through the process of upgrading the web app to build with Node 16 partly to avoid the extra step of having to upgrade NPM separately after installing Node 14. We're also looking at using some Node 16-specific features in the near future, but those changes aren't in yet.

Would you mind changing this PR to recommend using Node 16 instead?

neallred:
Sure, sounds good to me!

hmhealey (approved):

cwarnermm (approved):
Thanks so much for this great contribution update, @neallred!

neallred:
You're welcome!

justinegeffen (approved):

PR #873, opened 2021-08-04
Clarify the use of empty forms for no arguments commands

larkox:
#### Summary
Clarify the use of empty forms for no arguments commands

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35095


larkox:
What do you mean by fix? If you talk about the problem with "having to provide an empty list of fields" that is already solved on master. Sending an empty form works just fine at least on the test app.

If you mean a more long-term solution (different calls for each action, etc...) we would still need a design or decision on how to tackle that.

levb:
1. If a `call` to a `/form` returns a 404, the proxy can put an empty form in (HACK ALERT)
2. The root fix goes to untying the 3 URL, so that a binding can just have a submit and not a form, cc @mickmister.

justinegeffen (approved):

larkox:
Closing as we will change the Calls in several ways to avoid this problem.

PR #901, opened 2021-08-06
Update server-build.md

Sayanta66:
Fixed broken link


justinegeffen:
Thank you for this PR. The existing link is accurate and so this change will not be applied.

PR #916, opened 2021-08-15
Create new section for plugins: list of available plugins

justinegeffen:
This content was originally in the product docs and has been moved to the dev docs for ease of reference and also ease of update going forward as all the content will be in one place and not spread across doc sets. The content was copied as-is and no changes have been made. There may well be new content to add. 

@BenLloydPearson, the landing page for this section is a bit sparse and I wanted to find out if you have any content you'd like to add? I'm also open to having the sub-sections moved around. The only caveat is that I need to be able to redirect to them from the product docs site. 

Let me know if you have any questions!


BenLloydPearson:
In my opinion, the developer docs should focus on how to contribute code to Mattermost or build plugins and integrations. This content seems out of place to me in the developer docs because it's focused on consuming the plugins instead. I could buy the argument that we should have contribution guides for most of these since they are open source too, but I don't think this content fits that description.  Is that the intention with moving them to the developer docs?

justinegeffen:
Great questions. The content originally lived in the product docs in the same format. With the restructure of the content I thought it would work better to have all the plugin/extension/integration content in the same place. As all the integrations content is now on developers.mattermost.com this seemed like a logical place. 

I can think of a few options available:
1. Put these back in the product docs as part of how to extend Mattermost.
2. Remove them completely and rely on the Marketplace as a place for users to find plugins.
3. Add a minor section to the plugin/extend docs about the different categories of plugins available with a link to the Marketplace.

@aaronrothschild, do you have any other options in mind?

@BenLloydPearson, the above aren't the only options and I am 0/5 on the next steps especially as I want to enhance the work you've done!

BenLloydPearson:
I personally would prefer option 3. I think we should rely on the Marketplace page to surface this information and reference it wherever necessary. More improvements will be coming to the Marketplace page both during and after the v6 launch which should improve its ability to support this solution.

justinegeffen:
I'm happy to go with option 3 especially as it then provides a single source of truth that doesn't need constant updating if anything changes. I'll put together some content to replace these changes.

justinegeffen:
@BenLloydPearson, I'd like to keep the content for reference and in case we need it for some reason - would you be opposed to an "archive" directory being created here? I've provided a bit more content on the landing page but very open to additional input.

hanzei (commented):
This content seems to duplicate https://mattermost.com/marketplace/. Does it make sense to unify the content of the two sources?

justinegeffen:
Brilliant - thanks @hanzei! I'll go ahead and deprecate this content (delete it).

justinegeffen:
@aaronrothschild, @hanzei, I think this is good to approve and merge.

hanzei (changes requested):
1. The whole content looks a bit off. It reads more like a end-user documentation then developer focuses docs. What is the intention of placing the content here?

2. The headline indicates that this is some kind of general information page, while it's solely about the Marketplace. Should the name make that clear?

justinegeffen:
I don't think this is relevant anymore with the re-org of the pages.

PR #967, opened 2021-11-18
Add project_id to Android sign script

enahum:
#### Summary
Added project_id to be replaced in the strings resources as is needed by Firebase Cloud Messaging to be able to initialize and retrieve the device token to receive notifications.


sadohert:
Okay, test checks out.

Only question I have is how do we handle the fact that the `google-services.json` file I get from Firebase is missing the `firebase_url` property... 

I can't see a way for firebase to provide this?  Is it possible we could just form the `firebase_database_url` property we inject into `string.xml` ourselves, so customers dont' have to add it?  Does our app code care about this property, or is it in the underlying SDKs/etc.?

@enahum

enahum:
That is probably a good idea

sadohert:
I will take a stab at this.  I'm still puzzled about why we don't get the `firebase_url` property in `google-services.json`.  I mentioned in an [SO comment](https://stackoverflow.com/questions/66776899/cant-get-firebase-url-in-google-services-json/66777911#66777911).

Although, now that I think about it, if the full end to end build process still works then maybe the Google Services Maven plugin has already been updated to handle this 🤷🏼‍♂️?

site/static/scripts/sign-android
```diff
@@ -153,13 +153,16 @@ if [ "$PACKAGE_ID" != "" ]; then
     if [ "$GOOGLE_SERVICES_PATH" != "" ]; then
         echo "Replacing Google Services values"
         GOOGLE_SERVICES=$(jq . "$GOOGLE_SERVICES_PATH")
+        PROJECT_ID=$(echo "$GOOGLE_SERVICES" | jq -r .project_info.project_id)
         PROJECT_NUMBER=$(echo "$GOOGLE_SERVICES" | jq -r .project_info.project_number)
-        FIREBASE_URL=$(echo "$GOOGLE_SERVICES" | jq -r .project_info.firebase_url)
+        FIREBASE_URL="https://$PROJECT_ID.firebaseio.com"
+        echo "Setting  FIREBASE_URL to: $FIREBASE_URL"
```
enahum, line 159: @sadohert maybe remove this echo ?

sadohert: I left the echo in as a bit of a breadcrumb in case the logic picked something up incorrectly.

cwarnermm:
@sadohert - Friendly ping on this PR. Please note that a second reviewer is required for this repository.

sadohert:
Hey - Apologies..  I didnt' know I still had to do something here.

@enahum  how relevant is all this with Gekidou?

cwarnermm:
@sadohert - Friendly reminder to review this PR. Thanks!

cwarnermm:
@enahum @sadohert - Do we want to keep this PR open?

sadohert:
@cwarnermm  I'm not likely to get to this for a while, but I also don't want to lose the change if it comes up again with another customer (hard to know when this will rear its head).... is there a place this PR can sit that isn't in the way, but isn't gone for good?

cwarnermm:
Happy to leave it right here and open, @sadohert.

enahum:
closing, if ever needed we can re-open

PR #994, opened 2022-01-26
Retry of emdecr/reskin - Rebrand Alignment

emdecr:
#### Summary
This is a retry PR for #962.
When we merged and pushed to prod, the whole site went wonky.
Even weirder, the preview sites that were rendering before broke after we had already made reviews.

We thought propagation was at play, but that doesn't seem to be it.
Caching, as mentioned, could be at work.

I also capped this before we rolled back, it's looking for the bootstrap stylesheet.
Screen Shot 2022-01-21 at 4 46 50 PM

Thanks,
E

PR #1003, opened 2022-01-28
Update nav blog link to point to mattermost.com engineering blog

jwilander:
#### Summary
I noticed the blog link in the header was still pointing to the old developer blog instead of the blog on mattermost.com. This PR updates it to point to https://mattermost.com/blog/category/engineering/. I assume we'd rather drive traffic there instead of the old blog.

### DIFF
--- a/site/config.toml
+++ b/site/config.toml
@@ -77,7 +77,7 @@ pygmentsStyle = "manni"
       weight = 2
 
     [[menu.postpend]]
-      url = "/blog/"
+      url = "https://mattermost.com/blog/category/engineering/"
       name = "Blog"
       weight = 4
 


cwarnermm (approved):
Thanks, @jwilander!

emdecr (approved):
Thanks @jwilander – I'm up for the change to the site as it exists now. The preview site is looking broken for me, but I assume that's just because the site is at a subdir and the code is using absolute paths vs relative to pull in assets. The content change shouldn't be a breaking change.

With the rebrand branch #994 - we're linking to the more general /blog, but I look to @BenLloydPearson for guidance re: using /blog vs /blog/category/engineering/.

For the rebrand, the original idea was to mimic the nav on mm.com to a tee, but maybe given the presumed audience of this site that might be a good change? Not sure.

jwilander:
Thanks @emdecr. Looks like with the rebrand changes this doesn't apply any more so I'll close it out.

PR #1030, opened 2022-03-14
Remove mention of GOPATH

agnivade:
This is not necessary anymore. GOBIN is now automatically set from
the Makefile.


cwarnermm (approved):
Thank you, @agnivade!!

PR #1042, opened 2022-03-17
Update guilds with latest information

jwilander:
#### Summary
Remove some guilds that don't exist anymore, update some guild leads.


zefhemel (approved):
Yes, this is accurate (at least for web, server and mobile)

PR #1043, opened 2022-03-17
Update dev talks page to point to the youtube playlist

jwilander:
#### Summary
There were some old scripts meant for building the dev talks page that haven't worked in a while so I figured it's better to just link to the youtube playlist directly and remove the unneeded complexity.


lieut-data (approved):
:tada:

cwarnermm (approved):
Thanks!

PR #1064, opened 2022-05-04
Updates to web app developer docs

hmhealey:
This review ended up being a bit more involved than expected. Even ignoring the more nitpicky changes I made, some of the information around the web app process was wildly out of date. This ranged from out of date stuff that still shows up in the code like the non-TypeScript examples to things that were removed years ago like the references to Flux.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-43975


saturninoabril (approved):
Great, thanks @hmhealey!

hanzei:
@hmhealey I'm not the right person to review this. Maybe @michelengelen could jump in?

site/content/contribute/webapp/build-component.md
```diff
@@ -23,89 +23,92 @@ The components for the webapp currently live in the `components/` directory in t
...
-    */
-    items: PropTypes.arrayOf(PropTypes.object).isRequired
+     * An array of items to display
+     */
+    items: ListItem[];
 }
 ```
 
-The `title` prop would control the string to display for the title, while `items` would be an array of the item objects we would like to display. Note that `items` has `.isRequired` appended while `title` does not. This means that to render our component it must have the `items` prop set, while it can still render if we don't set `title`. Use this to provide optional props.
+The `title` prop would control the string to display for the title, while `items` would be an array of the item objects we would like to display. Note that `items` is required while `title` is optional.
```
michelengelen, line 50: not exactly sure if this is correct (from my end), but it sounds as if it should be without the `for`

```suggestion
The `title` prop would control the string to display the title, while `items` would be an array of the item objects we would like to display. Note that `items` is required while `title` is optional.
```

justinegeffen (approved):
Just one minor comment from me, otherwise great job! :)

site/content/contribute/webapp/build-component.md
```diff
@@ -114,60 +117,50 @@ export default connect(mapStateToProps, mapDispatchToProps)(ItemList);
...
 
-## Implementing your Component
+## Implementing the Component
 
-With the props defined and, if necessary, the container built you're ready to implement your component. For the most part, implementing a component for the webapp is no different than building any other React component. That said, we do have a few rules to abide by:
-
-1. All components must extend `React.PureComponent`.
-2. All data for rendering must come from props.
-3. Local rendering information that does not affect the store may be kept in state.
+With the props defined and, if necessary, the container built, you're ready to implement the rest of your component. For the most part, implementing a component for the web app is no different than building any other React component. While older code tends to use class components which inherit from `React.PureComponent`, most newer code should use functional components.
```
michelengelen, line 124:
```suggestion
With the props defined and, if necessary, the container built, you're ready to implement the rest of your component. For the most part, implementing a component for the web app is no different than building any other React component. While older code tends to use class components which extend from `React.PureComponent`, most newer code should use functional components.
```

site/content/contribute/webapp/unit-testing.md
```diff
@@ -14,43 +14,37 @@ The last required piece of building a webapp component is to test it. That can b
...
-        expect(wrapper.instance().renderOption()).toMatchInlineSnapshot();
-
-        // Only use "toMatchSnapshot" whenever above options are not possible.
-        // Limit the use to one (1) snapshot only.
-        // Save snapshot if it generates an easy to inspect and identifiable HTML or components that can easily verify future change.
         expect(wrapper).toMatchSnapshot();
+
+        // Save a snapshot of part of a component when it has another render function like "renderOption".
+        // This creates a small snapshot of that particular render function's result instead of the entire component.
+        expect(wrapper.instance().renderOption()).toMatchSnapshot();
```
michelengelen, line 31: this would only work with class-based components .. should we mention that here?

site/content/contribute/webapp/unit-testing.md
```diff
@@ -63,205 +57,169 @@ Below is a brief guide on how to do component testing:
...
             ...original,
             setGlobalItem: (...args) => ({type: 'MOCK_SET_GLOBAL_ITEM', args}),
         };
     });
     ```
 
-### 3. Getting "UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'data' of undefined"
-a. Use async mock function with resolved value. "Cannot read property" can be ``error``, ``data``, ``exists``, ``match``, or whatever the resolve value that the function returns.
+### 3. If you get an error like "UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'data' of undefined"
+a. Use async mock functions with resolved value. The property that cannot be readcan be `error`, `data`, `exists`, `match`, or whatever the resolved value contains.
```
michelengelen, line 182:
```suggestion
a. Use async mock functions with resolved value. The property that cannot be read can be `error`, `data`, `exists`, `match`, or whatever the resolved value contains.
```

michelengelen (approved):
just some smallish suggestions and minor typos. Nothing blocking as far as i can see! 🎉

site/content/contribute/webapp/build-component.md
hmhealey: You're right that it sounds a bit weird with the for, but I think it's similarly weird without it. I think that's caused by the "would"s in that sentence. How does this sound instead?
```suggestion
The `title` prop is a string that's displayed as the title of the list, and `items` is an array of objects that make up the contents of the list. Note that `items` is required while `title` is optional.
```

site/content/contribute/webapp/build-component.md
hmhealey: Just a slight tweak here since I don't think the "from" is needed if we switch to "extend"
```suggestion
With the props defined and, if necessary, the container built, you're ready to implement the rest of your component. For the most part, implementing a component for the web app is no different than building any other React component. While older code tends to use class components which extend `React.PureComponent`, most newer code should use functional components.
```

site/content/contribute/webapp/build-component.md
```diff
@@ -114,60 +117,50 @@ export default connect(mapStateToProps, mapDispatchToProps)(ItemList);
...
-            <div className='backstage-header'>
-                <h1>
-                    {this.props.title}
-                </h1>
-                {items}
-            </div>
-        );
-    }
+export default function ItemList(props: Props) {
+    const title = this.props.title ? <h1>{this.props.title}</h1> : undefined;
```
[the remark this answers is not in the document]
hmhealey: I've started favouring `undefined` over null since it makes typing cleaner in some situations like when typing function arguments since it avoids some awful type signatures like `Channel | undefined | null` that make me die a little inside. Ideally, JavaScript just wouldn't have two zero values for references, but this is what we're stuck with.

Ranting about JS aside, yeah, I can change this. Most people are probably more familiar with just using null everywhere, and that's better since this is an introductory doc
```suggestion
    const title = this.props.title ? <h1>{this.props.title}</h1> : null;
```

site/content/contribute/webapp/unit-testing.md
hmhealey: I don't think we need to mention it here. If it's a functional component, I'd just have that method into another component which could be tested independently, so this example isn't really relevant for them

site/content/contribute/webapp/build-component.md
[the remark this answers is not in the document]
michelengelen: haha, i feel ya ... there is a of discussion in the community at all times about the `null` vs `undefined` problem. I prefer to see it as if I define a variable manually I use `null`, since there is some intent in it.

PR #1101, opened 2022-08-16
Remove golang-webhook page

neflyte:
#### Summary

The existing “Using Mattermost with Webhook” page, which references a third-party product named [Webhook](https://github.com/adnanh/webhook) instead of a Mattermost Webhook, has been removed. Links to the removed page are redirected to `/integrate/webhooks`.


cwarnermm (approved):
Thanks, @neflyte!

azigler (approved):
I agree that removing this confusingly-named page is the best route, based on recent traffic.

PR #1130, opened 2022-10-04
Bump version to Mattermost 7.3.0

hanzei:
#### Summary
Bump Mattermost Server version to pull the latest plugin api docs.

The only updated page is `extend/plugins/server/reference/`.


go.mod
```diff
@@ -1,69 +1,69 @@
 module mattermost-developer-documentation
 
-go 1.17
+go 1.18
 
 require (
-	github.com/mattermost/mattermost-server/v6 v6.3.0
+	// mmgoget: github.com/mattermost/mattermost-server/v6@v7.3.0 is replaced by -> github.com/mattermost/mattermost-server/v6@5d1fe271e6
+	github.com/mattermost/mattermost-server/v6 v6.0.0-20220914120725-5d1fe271e68c
```
hanzei, line 7: See https://community.mattermost.com/core/pl/xq1jy5pw8ibrmpx9uwswusozey for more context on why a commit is used instead of a version.

neflyte:
The `go.mod` changes look okay and the docs build without error.
The changes to the build pipeline cannot be tested in a branch other than `master`, so we won't know if the changes work until this PR merges.

neflyte (approved):

neflyte:
Adding @spirosoik to review the build pipeline changes.

.gitlab-ci.yml
```diff
@@ -3,7 +3,7 @@
 variables:
   AWS_DEFAULT_REGION: "us-east-1"
 
-  IMAGE_BUILD: cimg/go:1.17-node
+  IMAGE_BUILD: cimg/go:1.18-node
```
spirosoik, line 6: it is a good practice to pin to a specific digest sha of the image build

spirosoik:
@neflyte for future reference just use @mattermost/core-build-engineers so someone can see it fast enough

.gitlab-ci.yml
hanzei: Is it really in this case? Go only releases security updates as patch releases, so we want to have them every time.

spirosoik: Not super critical for this repo but keep in mind that if we pin the dependencies to an exact version in our manifest files, no matter when you build your code, the same top-level dependencies should be pulled in, making our builds more transparent and consistent. This helps a bit in terms of the supply chain security

spirosoik (approved):

PR #1164, opened 2022-10-25
Add site search capability

neflyte:
### Summary

This PR adds basic site search functionality similar to that of the product docs site. A search box will be visible on the homepage of the docs site as well as just above the document links in the sidebar.

Search results will be displayed similarly to the product docs site, including keyword highlighting.

#### Hugo

Two layouts were added:

- `layouts/_default/search.html`: The search results page
- `layouts/_default/searchindex.json`: The search index data file

Two new content pages were added:

- `content/search/_index.md`: Creates the search result page using the `search` layout
- `content/search/searchindex.md`: Creates the search index data file using the `searchindex` layout

Two JavaScript files were added:

- `static/js/search.js`: Handles loading the search index and performing the search
- `static/js/search-page.js`: Initiates a search from a page other than the search results page

CSS styles for site search were added to the existing `static/css/styles.css` stylesheet.

##### Page URLs

The search results page is accessed at `/search/index.html` or `/search/`.
The search index data file can be accessed at `/search/searchindex/index.json`.

To perform a search via a URL, append a query parameter named `q` with the desired search terms as the value. For example:

`https://developers.mattermost.com/search/?q=app+manifest`

#### Lunr

The [Lunr](https://lunrjs.com) engine was selected for the site search. It is easy to implement, has a low size footprint, and several examples are available for integrating into Hugo.

Lunr supports some basic search keywords like `+` to include a keyword, `-` to exclude a keyword, and `~` for a fuzzy match. Further information on searching with Lunr can be found here: [Searching](https://lunrjs.com/guides/searching.html)

#### Field boosts

Since multiple fields are being searched, it becomes possible to weight search results differently depending on the fields that are matched. The Lunr search engine calls these weights “boosts”. If a matched field is configured with a boost, the score of the match is adjusted based on the configured value.

In this PR, the following field boosts are applied to search results:

- `title` +0.5
- `tags` +0.5
- `categories` +0.5
- `contents` +1.0
- `section` +0.5
- `subsection` +1.0

Field boosts can be configured in the `LunrSearch.buildIndex()` method of the `static/js/search.js` script.

For more information on Lunr boosts, see the [Boosts](https://lunrjs.com/guides/searching.html#boosts) section of the Lunr docs.


azigler (changes requested):
Looks cool @neflyte, I have some initial feedback for you:

- Is there a reason we're using lunr.js instead of something like Algolia? I keep going back to the [expo.dev documentation](https://docs.expo.dev/workflow/expo-cli/) because their search is so snappy and satisfying to use. Was anything else considered?
- Some other features I think would be useful are keyword highlighting, search-suggestions-as-you-type, and actually showing the relevant text snippet on the page in the search results (right now it's just a list of titles and their starting text).
- We probably also need a search landing page, or something else they see if they try to search an empty search bar. Right now it just takes them to an empty results page.
- We also need mock-ups again. Whenever a new visual component is added to the website, we need to have JuliAnn make a mock-up so we stay correctly branded. @cwarnermm is able to make this kind of ticket on Asana, or I can make it on behalf of y'all! To get the ticket made, can you make sure we account for all the new visuals? From my perspective, we need: the search bar on the doc landing page, the search bar on the top left of the TOC on docs, the little "loading..." toast, and the search results page. Anything else? If you confirm, I'll make the ticket!
- Can we make sure the entire search interface has appropriate aria labels for screenreaders?

I'm adding @emdecr as a potential reviewer in case she wants to double-check the accessibility, but no pressure there.

BenLloydPearson:
I'll echo Andrew's sentiments about Algolia. It would be awesome if we could incorporate that into our site.

Here are some additional points of feedback about this feature.
* More site hierarchy in the search results would be ideal. It's not clear where the search result is within the docs.
* Some search terms return multiple pages with the same title (e.g. "webhooks"), fixing the previous point could help with this, but maybe we need to take additional steps to fix it.
* Major landing pages are being buried under niche resources. For example, the only search phrase I found that led me to the Apps documentation was "app framework." All other terms I tried sent me to unrelated locations. It was similarly difficult to find search terms that surfaced the plugins and integrate landing pages. Is it possible to add weight to more important resources?

neflyte:
@azigler @BenLloydPearson Unfortunately, I took the path that was most compatible (and documented) for the Hugo engine we're using.

Also, Algolia is not free unless Mattermost is okay with only allowing 10,000 search requests per month. Lunr.js is 100% free and client-site, which matches the search on the product docs site.

It sounds like this functionality should be shelved for the time being; there seems to be more discussion needed.

I will move this PR back to draft.

azigler:
Hey @neflyte, [here are JuliAnn's mockups for search](https://www.figma.com/file/6C0zfdTL4RMlzBUXwalAQ9/Developer-Site?node-id=0%3A1&t=HWby2euvbXtD8lZh-0). Let me know if you have any questions!

azigler:
> @azigler The designs don't line up with what we've got in product w.r.t. the homepage. There are only 2 cards on the current homepage, down from 3 a few months ago. Could we get the designs updated today so that they will match production when this PR is merged?

Hey Alan! I can let JuliAnn know about that change on her template so it can be updated moving forward, but that shouldn't interfere with the PR being merged and matching the design. Thanks!

neflyte:
I've implemented as much of the design as I can without further input from JuliAnn; I've left a few comments/questions in the Figma doc for review. :smile:

azigler:
@neflyte Hey Alan, JuliAnn let me know you pointed out some implementation limitations about her mockup. Since Hugo won't support the sidebar, just exclude it for now and we can consider making a new issue for it. Also could you clarify what you mean by this:

> Please note that the product docs site uses the "Read the docs" site template while the dev docs site does not; it uses a completely different and custom template.

Are you saying the mockup should reflect a different theme?

cwarnermm:
@azigler - @neflyte's notes regarding the Figma design are intended to note that while the two docs sites (product & dev) are intended to behave similarly, the underlying tech for each site differs significantly. The product docs site currently uses a [Read the Docs theme](https://sphinx-rtd-theme.readthedocs.io/en/stable/) that was crafted specifically for Sphinx HTML generation, and it's that theme that controls the left-hand navigation, the main site elements, and the search field. 

As @neflyte dug deeper into the branding requests for the dev docs site, he realized that to implement the same search functionality, he'd need to port the Read the Docs layout code controlling that functionality to the dev docs templates via Hugo. This is a sizeable lift that may not offer enough value for the effort required. 

@neflyte applied a best effort approach to as many of the branding elements as he could given that both Sphinx and Hugo are likely not the tools we'll use long term to build and maintain these sites.

cwarnermm (approved):
Thank you, @neflyte, for this important development documentation update!

cwarnermm:
@asatkinson, @azigler, @emdecr - I've kicked off a fresh preview for this PR so that you can try out the new search mechanism @neflyte has added to the developer documentation. It should be ready soon.

My goal is to commit this PR to the codebase and make the search available in production where we can start gathering user feedback on its implementation and iterate on in future releases. 

Please note any blockers you encounter that must be addressed before merging. Please also note that @neflyte has transitioned to a part-time contribution role with Mattermost and as such, his availability is considerably reduced for the time being.

emdecr (commented):
Looks great, thanks @neflyte!

A couple comments:

1. I tend to see search inputs in a `form` element with the role attribute set to search - `role="search"`. From looking at a few resources ([1](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/search_role), [2](https://www.w3.org/WAI/tutorials/forms/instructions/)) that still seems to be accepted re: accessibility, but also have seen people put the role attribute on a `div` wrapper - cc: @azigler for any comments. I'd also suggest to keep some text within the `label`. I _think_ we can wrap the `input` with the `label` and add a visually hidden `span`, so that the text can still be picked up. If we use the `form` tag, we'd need to add a snippet of JS to prevent the default action, which would end up overriding the Lunr search functionality. I'm no expert, so open to ideas of course.

Back-of-the-napkin edits:

``` html
<!-- in hero.html -->
<div class="homepage-search__input">
    <form role="search" class="homepage-search__form" id="search-form">
        <label for="search-query">
            <span class="visually-hidden">Search</span>
            <input id="search-query" type="text" class="transparent ff--trade-gothic" placeholder="What are you looking for?">
        </label>
    </form>
</div>
```

``` css
/* in styles.css */
.visually-hidden {
    border: 0;
    clip: rect(0 0 0 0);
    -webkit-clip-path: inset(50%);
    clip-path: inset(50%);
    width: 1px;
    height: 1px;
    margin: -1px;
    overflow: hidden;
    padding: 0;
    position: absolute;
    white-space: nowrap;
}
```

``` js
// in main.js
const searchForm = document.getElementById('search-form');
searchForm.addEventListener('submit', (e) => {
    e.preventDefault();
});
```
2. Also, when I end up on the search results page, I can't seem to enter a new search term and do a new search? Is that a dev/preview site glitch?

If we've got less bandwidth and need a hand, I can also try to push the structural/styling changes up to this PR branch. Obvs no pressure though.

Lemme know what you think - thanks All!

cwarnermm:
Thank you, @emdecr, for this great feedback! I'd welcome your assistance with these updates if you have the bandwidth!

azigler:
Thanks for the ping and for auditing the PR, @emdecr! @neflyte @cwarnermm I went ahead and apply the a11y changes. In this case, we didn't need a `<label>` tag because we can label the form as `role="search"` and then the actual `<input>` as `type="search"` and add an `aria-label` to the `<input>` directly. This obviates the need for the `<label>` element or hiding CSS and is on-spec. I also adjusted the pre-existing CSS a little. Ex:

```html
<form role="search" class="search__form" id="search-form">
    <div class="search__input">
        <input id="search-query" class="search__query-input ff--trade-gothic" type="search" placeholder="What are you looking for?" aria-label="search">
    </div>
</form>
```

I applied the JS you suggested to prevent the default form submission, and the search works correctly on all the pages except the search results page. On that page, I get this devtools console error when I press ENTER to search:

```
search.js:141 Uncaught TypeError: this.search is not a function
    at HTMLInputElement.searchInputWasSubmitted (search.js:141:18)
```

@neflyte if you can troubleshoot that and get it working again, I think this PR will be working and ready to go.

neflyte:
@azigler I fixed up a piece of code that handled the <Enter> keypress so it calls the correct search function instead of throwing the error you saw. :slightly_smiling_face: For the life of me I don't know why this worked on my machine 2 weeks ago and not today.

cwarnermm:
@azigler - A fresh preview will be available shortly to test. Thanks!

azigler:
@neflyte @cwarnermm Looks like it's working now! My final suggestion is to make it so that when searching on the search page, the query param in the URL is updated onEnter to match the latest query. Otherwise, it sticks as the query used when the user first lands on the page. If they search something else and get new results then if they copy-paste the URL it will be wrong, for example.

asatkinson (approved):
I tried the search functionality on a number of terms and I got equivalent or better results than doing so in Google search which seems like a good test.
I think users will generally be pleased with the results.
At some point it would be nice to combine with the rest api search.

I didn't review the technical implementation - I'll leave that to others to discuss.

neflyte:
@azigler I made a change that makes the search field on the results page redirect back to the same page with the new query. This will ensure that the URL query parameters match the value of the search field on the results page when a new search is executed. :smile:

cwarnermm:
@azigler - A fresh preview is being generated now and should be available shortly for your review.

azigler (approved):
Thanks for squashing that last bug, @neflyte!

cc @cwarnermm

PR #1165, opened 2022-10-27
Removed $ signs so the commands can be copied with the UI copy button

javaguirre:
#### Summary

Right now the `$` sign was copied from the highlighted command in the UI, so when you want to execute it won't work right away.


cpoile (approved):
Looks good, thanks @javaguirre !

cwarnermm (approved):
Thank you, @javaguirre!

PR #1167, opened 2022-10-31
Fix broken builds; webapp registry.js migrated to TypeScript

neflyte:
#### Summary
The `mattermost-webapp/master/plugins/registry.js` file was converted to TypeScript and renamed to `registry.ts`. This breaks the dev docs build since it's expecting the previous filename and that it contains ECMAScript.

The ECMAScript parsing was replaced by TypeScript parsing by way of `@typescript-eslint/typescript-estree`, which is a mostly drop-in replacement for `espree`. The logic in `plugin-jsdocs.js` was updated to extract function arguments from new locations in the abstract source tree (AST).

### DIFF
--- a/scripts/package-lock.json
+++ b/scripts/package-lock.json
@@ -9,27 +9,110 @@
       "version": "0.0.1",
       "license": "MIT",
       "dependencies": {
-        "espree": "^7.0.0",
+        "@typescript-eslint/typescript-estree": "5.41.0",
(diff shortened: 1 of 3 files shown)


cwarnermm (approved):
Thank you for identifying a quick fix for this issue, @neflyte! I'll defer to feedback provided through technical reviews.

hmhealey (approved):

PR #1169, opened 2022-11-07
Update incoming webhook response parameters

neflyte:
#### Summary

This PR fixes incorrect information about responses to Incoming Webhooks with regard to the `username` and `icon_url` properties. Contrary to the existing docs, the `username` and `icon_url` properties will not inherit their values from the user who created the incoming webhook.


azigler (commented):
Looks good to me, if this is the case. Do these potentially have different values depending on if `Enable integrations to override usernames` et al are enabled via the System Console?

jprusch:
It should be mentioned, that if no username / icon is set in the webhook configuration dialog, it defaults to "webhook" and the default webhook image.

levb (approved):

cwarnermm (approved):
Thanks, @neflyte!

neflyte:
@azigler Yes, they will default to the name `webhook` and the standard webhook icon. The doc doesn't call this out clearly, though. :slightly_smiling_face: I'll fix that up!

neflyte:
@jprusch Thanks! I will add that detail. :slightly_smiling_face:

neflyte:
I made similar changes to the Outgoing Webhooks page as well. :)

azigler (approved):
Thanks for clarifying @neflyte!

PR #1180, opened 2022-12-05
Add missing feedback widget to the rest of the docs pages

neflyte:
#### Summary
The feedback widget is missing from several docs pages. This PR resolves the issue by adding the feedback widget code to the docs template that was not modified during the original implementation.


cwarnermm (approved):
Thanks, @neflyte!

azigler (approved):
Looks good!

PR #1192, opened 2022-12-30
Update Hugo version to 0.101.0 for production builds

neflyte:
#### Summary

The version of Hugo used to build the docs in production is still using v0.92.1 and may be the cause of some rendering errors. This PR updates the version of Hugo to v0.101.0, so it matches the version used in GitHub actions.


cwarnermm (approved):
Thanks, @neflyte!

justinegeffen (approved):

PR #1195, opened 2023-01-13
Correct typo in `docker-compose` command

oh6hay:
#### Summary

Corrects a typo in the command to start the docker-compose stack. `docker compose up` is not a valid Docker command, meanwhile `docker-compose up` is a valid `docker-compose` command.

#### Ticket Link
n/a


oh6hay:
Hmm, the documentation consistently mentions `docker compose` as a valid command so I checked the `docker` command on my mac. Seems some `docker` clients contain a `compose` command, some do not. Not sure it makes sense to more explicitly refer to `docker-compose` in a manner that works on all platforms, as the likely audience can figure this out anyway.

PR #1220, opened 2023-03-01
Add documentation around the Show Logs function

devinbinnie:
#### Summary
We added a menu item called 'Show Logs' that points out where the log file is location for ease of use.


cwarnermm (approved):
Thanks, @devinbinnie!

m1lt0n (approved):

PR #1226, opened 2023-03-14
Remove underscore from action id

anx-ag:
According to https://mattermost.atlassian.net/browse/MM-49350, underscores are not allowed in action ids. I stumbled upon that some time ago and a [user in the forums](https://forum.mattermost.com/t/interactive-message-menu-not-returning-value/10841/10) brought this issue back to my attention, so here's a fix for the documentation, at least to make sure the docs have a hint about that and users (like me) don't copy a broken example config.

### DIFF
--- a/site/content/integrate/plugins/interactive-messages/_index.md
+++ b/site/content/integrate/plugins/interactive-messages/_index.md
@@ -120,7 +120,7 @@ Similar to buttons, add message menus as `actions` in your integration [message 
 
 ![image](message-menus.png)
 
-The following payload gives an example that uses message menus.
+The following payload gives an example that uses message menus (please note that the id in the actions array may only consist of letters and numbers, no other characters are allowed!):
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


site/content/integrate/plugins/interactive-messages/_index.md
```diff
@@ -120,7 +120,7 @@ Similar to buttons, add message menus as `actions` in your integration [message
 
 ![image](message-menus.png)
 
-The following payload gives an example that uses message menus.
+The following payload gives an example that uses message menus (please note that the id in the actions array may only consist of letters and numbers, no other characters are allowed!):
```
cwarnermm, line 123:
```suggestion
The following payload gives an example that uses message menus (The id in the actions array may only consist of letters and numbers, no other characters are allowed):
```

cwarnermm (approved):
Thanks, @anx-ag!

PR #1260, opened 2023-06-29
Update developer-setup.md

jlandells:
The instruction was to clone `mattermost-server.got`, but the repo that was cloned is actually `mattermost.git`


sadohert:
Thanks for the fix @jlandells !  

This happened as a result of the recent transition to a "Monorepo" where we have brought together a few different Mattermost product repos into a single one.

crspeller (approved):

PR #1277, opened 2023-07-20
manual cherry pick doc: suggest a remote branch name for the push

JulienTant:
#### Summary
Small change to suggest using the same branch name when pushing the new branch. I didn't do that today and ended up somehow pushing my local branch directly to cloud 🤦🏻 

#### Ticket Link
N/A


cwarnermm (approved):
Thank you, @JulienTant!

lieut-data (approved):
Thanks, @JulienTant! This definitely seems like a place to be safer rather than sorry.

> Small change to suggest using the same branch name when pushing the new branch. I didn't do that today and ended up somehow pushing my local branch directly to cloud 🤦🏻

Interesting, I guess that's because we don't lock the `cloud` branch? Thoughts @toninis / @spirosoik?

PR #1287, opened 2023-08-16
Correct max post size on slash command page

robinsdm:
Correct the maximum post size documented on the slash commands page. The v5.0 release notes say it is 16383 characters.


cwarnermm (approved):
Thanks, @robinsdm!

mickmister (approved):

PR #1298, opened 2023-10-05
[fix] CI, fixing check-label-preview IF clause Scope

phoinixgrr:
Rationale: https://mattermost.atlassian.net/browse/CLD-5750

### DIFF
--- a/.github/workflows/check-label.yml
+++ b/.github/workflows/check-label.yml
@@ -7,10 +7,8 @@ on:
 jobs:
   check-label-preview-env:
     runs-on: ubuntu-latest
+    if: contains(github.event.pull_request.labels.*.name, 'preview-environment')
     steps:
-    - name: check-label
-      if: contains(github.event.pull_request.labels.*.name, 'preview-environment')
-      run: echo "label exist, will trigger the preview environment"
     - name: Set PR
       env:
         PR_NUMBER: ${{ github.event.pull_request.number }}


.github/workflows/check-label.yml
```diff
@@ -7,9 +7,9 @@ on:
 jobs:
   check-label-preview-env:
     runs-on: ubuntu-latest
+    if: contains(github.event.pull_request.labels.*.name, 'preview-environment')
     steps:
     - name: check-label
```
jupenur, line 12: Having this here makes very little sense now that the actual check is higher up. Maybe just remove it altogether?

phoinixgrr: @jupenur  In which actual check you are referring to?

jupenur: The `if` you moved away from under this step. The step now does nothing.

phoinixgrr: @jupenur  The step prints in STD OUT: "label exist, will trigger the preview environment"

phoinixgrr: Is that nothing? 
There is some verbosity introduced by the original author. Is there a reason for removing this?

jupenur: Well yes, but that doesn't seem useful. It does not "check the label" like the name suggests, and the job running should already tell anyone reading the logs that the label exists and the preview environment is getting triggered.

So what I'm saying is this step contains zero functionality: it just logs a line of text, and even that log entry is entirely redundant.

phoinixgrr: Rebased and pushed the changes asked.

jupenur (approved):
:shipit:

saturninoabril (approved):
Thanks @phoinixgrr!

PR #1301, opened 2023-10-13
[MM-54869] Add nvm as the default NodeJS provider for the Desktop App

devinbinnie:
#### Summary
`nvm` is an easier way to manage Node versions per machine, so I think it makes sense that we support it as our default way to install Node for the Desktop App as well. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-54869


cwarnermm (approved):
Thanks, @devinbinnie!!

hmhealey (approved):

PR #1311, opened 2023-11-03
Add prerequisites for WSL setup.

vish9812:
#### Summary

Add prerequisites for setting up Mattermost on a WSL environment.


site/content/contribute/developer-setup.md
```diff
@@ -19,8 +19,20 @@ If you're migrating from before the monorepo see the [migration notes]({{< ref "
 If you're developing plugins, see the plugin [developer setup]({{< ref "/integrate/plugins/developer-setup" >}}) documentation.
 {{</note>}}
 
+# Prerequisites for WSL Development Environment
+
+1. [Install WSL](https://learn.microsoft.com/en-us/windows/wsl/install) by running the following command as an administrator in PowerShell: `wsl --install`
+2. [Install Docker Desktop for Windows](https://learn.microsoft.com/en-us/windows/wsl/tutorials/wsl-containers#install-docker-desktop) on your Windows machine.
+3. Now make sure to perform the rest of the operations(except Docker installation) within the WSL environment and not in Windows.
```
cwarnermm, line 26:
```suggestion
3. Perform the rest of the operations (except Docker installation) within the WSL environment and not in Windows.
```

cwarnermm (approved):
Thanks so much, @vish9812! I've provided inline feedback. Approving.

site/content/contribute/developer-setup.md
```diff
@@ -19,8 +19,20 @@ If you're migrating from before the monorepo see the [migration notes]({{< ref "
 If you're developing plugins, see the plugin [developer setup]({{< ref "/integrate/plugins/developer-setup" >}}) documentation.
 {{</note>}}
 
+# Prerequisites for WSL Development Environment
+
+1. [Install WSL](https://learn.microsoft.com/en-us/windows/wsl/install) by running the following command as an administrator in PowerShell: `wsl --install`
+2. [Install Docker Desktop for Windows](https://learn.microsoft.com/en-us/windows/wsl/tutorials/wsl-containers#install-docker-desktop) on your Windows machine.
```
crspeller, line 25: I have had much better luck with installing Docker directly in the distribution after enabling systemd support: https://learn.microsoft.com/en-us/windows/wsl/wsl-config#systemd-support

Maybe we should suggest that instead or as well as docker desktop?

site/content/contribute/developer-setup.md
```diff
@@ -19,8 +19,20 @@ If you're migrating from before the monorepo see the [migration notes]({{< ref "
...
+3. Perform the rest of the operations (except Docker installation) within the WSL environment and not in Windows.
+
 # Setup the Mattermost Server
 
+1. Install build essential tools:
+
+   ```sh
+   sudo apt install build-essential
```
crspeller, line 33: This is Ubuntu specific. If we want to add this detail we should call this out. @cwarnermm is there an existing pattern for this we want to use? So far this doc is pretty generic.

crspeller (changes requested):

site/content/contribute/developer-setup.md
vish9812: From what I've experienced, devs on Windows generally have some kind of setup already on Windows which also includes the Docker Desktop. Also, the Docker Desktop plays nicely when you have multiple Linux setup as WSL. And it also _supports both_ Windows and WSL.
So, I would still like to keep the Docker Desktop as an option. But yeah, we can also add the "direct-docker" as an alternative.

site/content/contribute/developer-setup.md
vish9812: Maybe we can just say here to install "make". And then probably give this as an example like on Ubuntu you can do this?

site/content/contribute/developer-setup.md
crspeller: Sounds good.

site/content/contribute/developer-setup.md
crspeller: Makes sense.

cwarnermm (approved):

crspeller (approved):

PR #1326, opened 2024-01-08
Remove Google Summer of Code page

agarciamontoro:
#### Summary
This page has been there for a while, and it has created some confusion from time to time, letting contributors believe that Mattermost was participating in Google Summer of Code. As that is not the case, let's just remove it.

#### Ticket Link
--


hanzei:
:cry:

cwarnermm (approved):

Issue #1281, opened 2023-07-28
Help Wanted: OpenSSL 3.0 legacy provider note


jacob-morin:
Hi all, 

As of OpenSSL 3.0, some algorithms (in this case specifically, RC2) have been moved to the `legacy` provider, which is not loaded by default (see https://wiki.openssl.org/index.php/OpenSSL_3.0). This causes the command `openssl pkcs12 -in Certificates.p12 -out aps_production_priv.pem -nodes -clcerts -passin pass:` on line 92 of `site/content/contribute/more-info/mobile/push-notifications/service.md` to fail with the cryptic error: 
```
Error outputting keys and certificates
00E68F1701000000:error:0308010C:digital envelope routines:inner_evp_generic_fetch:unsupported:crypto/evp/evp_fetch.c:341:Global default library context, Algorithm (RC2-40-CBC : 0), Properties ()
```

I got around this issue by adding the `-legacy` option to my command, which loads the legacy provider (and therefore the RC2 algorithm). Perhaps this is worth noting in the documentation?

Omar8345:
Can I get assigned to this?

cwarnermm:
Thank you, @Omar8345, for your help with this issue!

PR #1332, opened 2024-01-22
[GH-1281] OpenSSL Compatibility: Add -legacy option

Omar8345:
#### Summary

This pull request fixes a OpenSSL command in the [Install the Mattermost push notification service documentation](https://developers.mattermost.com/contribute/more-info/mobile/push-notifications/service/), which extracts the private key from a certificate, by adding the `-legacy` option, this is due to since OpenSSL 3.0, the RC2 algorithm has been moved to the `legacy` provider, which is not loaded by default.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-developer-documentation/issues/1281

### DIFF
--- a/site/content/contribute/more-info/mobile/push-notifications/service.md
+++ b/site/content/contribute/more-info/mobile/push-notifications/service.md
@@ -89,7 +89,7 @@ For the sake of making this guide simple we located the files at `/home/ubuntu/m
     ```
 - Extract the private key from the certificate you exported ..
     ```sh
-    $ openssl pkcs12 -in Certificates.p12 -out aps_production_priv.pem -nodes -clcerts -passin pass: -legacy
+    $ openssl pkcs12 -in Certificates.p12 -out aps_production_priv.pem -nodes -clcerts -passin pass: -legacy -rc2
     ```
 - Verify the certificate works with Apple
     ```sh
--- a/site/content/contribute/more-info/mobile/push-notifications/service.md
+++ b/site/content/contribute/more-info/mobile/push-notifications/service.md
@@ -91,7 +91,7 @@ For the sake of making this guide simple we located the files at `/home/ubuntu/m
     ```sh
     $ openssl pkcs12 -in Certificates.p12 -out aps_production_priv.pem -nodes -clcerts -passin pass: -legacy
     ```
-- Verify the certificate works with apple
+- Verify the certificate works with Apple
     ```sh
     $ openssl s_client -connect gateway.push.apple.com:2195 -cert aps_production.pem -key aps_production_priv.pem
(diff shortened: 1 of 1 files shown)


site/content/contribute/more-info/mobile/push-notifications/service.md
```diff
@@ -89,9 +89,9 @@ For the sake of making this guide simple we located the files at `/home/ubuntu/m
     ```
 - Extract the private key from the certificate you exported ..
     ```sh
-    $ openssl pkcs12 -in Certificates.p12 -out aps_production_priv.pem -nodes -clcerts -passin pass:
+    $ openssl pkcs12 -in Certificates.p12 -out aps_production_priv.pem -nodes -clcerts -passin pass: -legacy
```
fmartingr, line 92: Since using the `-legacy` flag changes the default algorithms, maybe is better if we also specify the algorithm we want in the command? To avoid problems for people using different OpenSSL versions thus potentially providing different defaults. What do you think?

esarafianou:
This issue stems out of the fact that Apple’s .p12 files use older encryption standards including TripleDES-CBC and RC2. Although I don't like suggesting our customers to use legacy algorithms and flags, there's probably no other option before Apple updates they encryption standards for .p12 files. But before I approve this, I'd like @jupenur's input if there's any better alternative

jupenur (approved):
The only potential issue I see here is that OpenSSL versions prior to 3.0 won't support the `-legacy` flag, meaning the command will fail. But as of last September that's not really super relevant any more either, since 1.1.1 EOL'd.

esarafianou (approved):

site/content/contribute/more-info/mobile/push-notifications/service.md
Omar8345: Good idea.

fmartingr (approved):
👍

Omar8345:
@cwarnermm The code review I bet it's done, you can just approve the review, well, nothing is there to be reviewed by you except capitalizing the first letter in the word Apple (company).

cwarnermm (approved):

tnir:
Thanks for this workaround for certificate-based connection, but it looks so old-fashioned. [Push proxy](https://github.com/mattermost/mattermost-push-proxy) should support token-based connections to APNs 😁 

https://developer.apple.com/documentation/usernotifications/establishing-a-token-based-connection-to-apns

> Token-based authentication offers a stateless way to communicate with APNs. Stateless communication is faster than certificate-based communication because it doesn’t require APNs to look up the certificate, or other information, related to your provider server. There are other advantages to using token-based authentication:

PR #1337, opened 2024-02-13
Updating the plugin version to generate the new documentation

jespino:
The plugin version is relevant to generate the documentation of the server API,
was using the version 0.0.5 and we are now in the 0.0.13, so we need to have
that to include all the new API endpoints and hooks.


cwarnermm (approved):
Thank you, @jespino! Is there a way to manage this as part of monthly releases moving forward? We've encountered this issue before, and it would be ideal to proactively address it rather than encounter missing content on the dev site.

go.mod
```diff
@@ -3,68 +3,48 @@ module mattermost-developer-documentation
 go 1.19
 
 require (
-	github.com/mattermost/mattermost/server/public v0.0.5
+	github.com/mattermost/mattermost/server/public v0.0.13
```
hanzei, line 6: Nit: There is already a `v0.0.14`

hanzei (approved):
Thanks for keeping the docs up to date :+1:

go.mod
jespino: Let me update that.

jespino:
@cwarnermm good point, I would love to have that some kind automatized, because if not is going to get again  out of date over a over again. Probably we can add something the CI. @saturninoabril do you know who would be the right person to add that CI check?

It would be something like checking if `go list -u -m github.com/mattermost/mattermost/server/public` output is contained in go.mod file.

agarciamontoro (approved):

hanzei:
Alternatively, we can configure Dependabot to update `go.mod` automatically.

PR #1350, opened 2024-03-22
Add graddle lockfile update to the documentation

larkox:
#### Summary
We add documentation on how to update the lock files.

This is dependent on https://github.com/mattermost/mattermost-mobile/pull/7875

#### Ticket Link
NONE

### DIFF
--- a/site/content/contribute/more-info/mobile/build-your-own/android.md
+++ b/site/content/contribute/more-info/mobile/build-your-own/android.md
@@ -128,3 +128,9 @@ We use lockfiles to lock dependencies and make sure the builds are reproductible
 cd android
 ./gradlew app:dependencies --update-locks "*:*"
 ```
+
+In case we want to regenerate the lockfile from the scratch, we can delete the `android/buildscript-gradle.lockfile` and then run the following commands:
+```
+cd android
+./gradlew app:dependencies --write-locks
+```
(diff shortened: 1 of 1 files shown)


site/content/contribute/more-info/mobile/build-your-own/android.md
```diff
@@ -120,3 +120,11 @@ $ npm run build:android
 ```
 
 This will start the build process following the environment variables you've set. Once it finishes, it will create the `.apk` file(s) with the `APP_NAME` as the filename in the project's root directory. If you have not set Fastlane to submit the app to the Play Store, you can use this file to manually publish and distribute the app.
+
+## Frequently Asked Questions
+### How do I update the lock file?
+We use lockfiles to lock dependencies and make sure the builds are reproductible. If we want to update the lockfile to update all dependencies to the latest, we can run these commands:
+```
+cd android
+./gradlew app:dependencies --update-locks "*:*"
```
enahum, line 129: What if we dont want the latests but the specific verdions we are using?

larkox: The specific versions we are using would be already in the repo (in the lockfile).

If we want to generate them from the scratch (remove the lock file and regenerate it) we would call `./gradlew app:dependencies --write-locks`. Is that what you mean?

enahum: Yes.. cause this is how I envision it.

Add a new dependency to the project or bump dependencies manually, then update the lock.. kind of the same we do with gems, cocoapods and npm

cwarnermm (approved):

enahum (approved):

cwarnermm (approved):

Issue #1378, opened 2024-06-10
Instruction for stopping docker container is given in developing Mattermost without docker documentation


chessmadridista:
# What is the issue?
In the section titled "Develop Mattermost without Docker" on [this link](https://developers.mattermost.com/contribute/developer-setup/#develop-mattermost-without-docker:~:text=the%20running%20docker%20containers%3A-,make%20stop%2Ddocker,-Set%20your%20options%3A%20Some), the code for stopping docker container is given.

If we are developing Mattermost without Docker, is there any need for that code snippet?

# Proposed solution

> The stop-server make target does not stop all the docker containers started by run-server. To stop the running docker containers:
> 
> ```
> make stop-docker
> ```
> 

This whole section can be removed for better readability.

chessmadridista:
I would be happy to take up this issue if it is something that needs to be resolved.

cwarnermm:
Thank you, @chessmadridista!

PR #1380, opened 2024-06-10
GH-1378

chessmadridista:
#### Summary

Docker code snippet along with text description in `develop without docker` section has been removed.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-developer-documentation/issues/1378

### DIFF
--- a/site/content/contribute/developer-setup/_index.md
+++ b/site/content/contribute/developer-setup/_index.md
@@ -178,10 +178,6 @@ If you're using Windows, we recommend using the Windows Subsystem for Linux (WSL
     ```sh
     make stop-server
     ```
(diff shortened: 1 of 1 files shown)


mickmister (approved):

hmhealey (approved):
Thanks for the PR, @chessmadridista. Great catch! It's far too easy to end up copying something like that when keeping two different sets of instructions in sync

chessmadridista:
- > ...It's far too easy to end up copying something like that when keeping two different sets of instructions in sync

  Yes, it is understandable.

- > Thanks for the PR, @chessmadridista. Great catch!...

  Thank you and always welcome! This is my first ever open source contribution so I am glad that I was able to help.

PR #1384, opened 2024-06-14
Update MVP

amyblais:


lindy65 (approved):
Thanks Amy, looks great! I think this is the first time ever we've had 2 MVPs for one release :)

amyblais:
> Thanks Amy, looks great! I think this is the first time ever we've had 2 MVPs for one release :)

There have been 2 winners in the past, but it happens rarely.

cwarnermm (approved):

PR #1396, opened 2024-09-05
Clarified that bot accounts can't be used to upload files to a channel

cwarnermm:
Based on customer feedback via @jlandells


jlandells (approved):
Perfect!

PR #1399, opened 2024-09-17
Update _index.md

jlandells:
Added more detail to when a bot is allowed to upload a file to a channel


cwarnermm (approved):
Thanks, @jlandells!!

site/content/integrate/reference/bot-accounts/_index.md
```diff
@@ -11,7 +11,7 @@ Bot accounts access the Mattermost {{< newtabref href="https://api.mattermost.co
 
   - Can't be logged into.
   - Can't be used to create other bot accounts.
-  - Can't be used to upload files to a channel.
+  - Can't be used to upload files to a channel, unless they are also added to the channel as a channel member.
```
hanzei, line 14: I don't understand why file upload is specially mentioned in this list. File upload works the same way as other API methods. What makes it special so we need to list it here?

hanzei (changes requested):

site/content/integrate/reference/bot-accounts/_index.md
jlandells: The fact that an important customer struggled with this is a good indication that we needed to improve the docs.  Consider that a bot can **post** to a channel without being a member, so it’s not intuitive that they need to be added to the channel in order to upload files.

jlandells: We need to get a resolution on this.  Either we remove the line altogether and leave it confusing for customers, we go with highlighting the area of confusion, or we need to rewrite the page entirely.  It may seem obvious to us, but if customers are wasting time trying to figure something out and a simple doc change (even one that we feel should be obvious) can help them, isn't that a valid change?

What do you think, @cwarnermm ?

cwarnermm: I'm 5/5 with @jlandells on this issue: if customers are struggling with a workflow, the docs are one excellent place to provide clear guidance towards success.

@hanzei - Would you be open to creating a docs PR with the changes you proposed above (re: step 7)?

hanzei: > I'm 5/5 with @jlandells on this issue: if customers are struggling with a workflow, the docs are one excellent place to provide clear guidance towards success.

I completely agree that the docs should provide guidance for customers on common problems.

My concern is that the list that we are editing starts with 

> Bot accounts are just like user accounts, except they:

And the statement that the PR adds is also true for normal users. The statement highlights something that isn't correct. That might cause additional confusion for customers.

Hence, I wonder if we should move the sentence to the FAQ section later in the document. I'm happy to drive that change.

hanzei (changes requested):
I am approving with concerns raised in my comment. I think we need to fine tune the page further.

hanzei (approved):
wrong button :sweat_smile:

site/content/integrate/reference/bot-accounts/_index.md
cwarnermm: You make really good points, @hanzei, and I appreciate your willingness to update this PR to move these guidance details to the FAQ section.

jlandells:
Too much back and forth for such a minor change.  Will leave it as is and let the customers figure it out. 😞

hanzei:
I don't understand why you closed the PR. This is an improvement, and we should merge it. As I pointed out in https://github.com/mattermost/mattermost-developer-documentation/pull/1399#discussion_r1796639861, we can make the docs even clearer, but that can be a follow-up.

jlandells:
I get it, but whilst this has sat open and we've been discussing it, it got out of sync and I'd have had to do additional work to add a simple note that there seemed to be contention around.  It felt better to just let the docs be cleaned up by someone who understands this better than I do.

hanzei:
I'm happy to bring the PR over the finish line. CI is green. Would you mind if I merge your change @jlandells ?

jlandells:
>  Would you mind if I merge your change @jlandells ?

Go for it! 😎

PR #1406, opened 2024-10-03
Replace deprecated command with environment file

jongwooo:
Update `.github/workflows/preview-env-fork.yml` to use environment file instead of deprecated `set-output` command. 
For more information, see: [https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)

I found the workflow file that use `set-output` command through the following command:

```bash
$ find . -name '*.yml' -o -name '*.yaml' | xargs egrep '\bset-output\b'
```

**AS-IS**

```yaml
echo "::set-output name=id::$(<pr.txt)"
```

**TO-BE**

```yaml
echo "id=$(<pr.txt)" >> $GITHUB_OUTPUT
```

### DIFF
--- a/.github/workflows/preview-env-fork.yml
+++ b/.github/workflows/preview-env-fork.yml
@@ -21,7 +21,7 @@ jobs:
       - name: Save PR number
         id: pr
         run: |
-          echo "::set-output name=id::$(<pr.txt)"
+          echo "id=$(<pr.txt)" >> $GITHUB_OUTPUT
       - uses: actions/checkout@v2
         with:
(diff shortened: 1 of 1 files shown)


toninis (approved):
🚀

cwarnermm:
Thank you, @jongwooo, for your Mattermost Developer Documentation contribution and for logging a dev docs issue in this area. While we won't be moving forward with your documentation PR, I can confirm that the issue you logged will be addressed in a future docs release by our Delivery team.

PR #1407, opened 2024-10-04
feat: Upgrade the way we deploy preview environments from forks

toninis:
#### Summary

Use the same foolproof mechanism we use on https://github.com/mattermost/docs repository to handle fork builds


mvitale1989 (approved):
LGTM, thanks Antonis! :rocket:

cwarnermm (approved):
Thank you, @toninis!!

PR #1413, opened 2024-11-01
Update package manager name for RHEL developer-setup

jonathan-dove:
#### Summary
Incorrect package manager was referenced as apt in the RHEL developer documentation. Simple change of apt to yum. Utilized yum instead of dnf as most distributions support either or but not all RHEL variants support dnf if someone was to utilize a EL7 variant.


devinbinnie (approved):
LGTM! Thanks @jonathan-dove!

cwarnermm (approved):
Thanks, @jonathan-dove!

PR #1424, opened 2024-12-05
Replace deinitialize with uninitialize

hmhealey:
#### Summary
Very early on, web app plugins had a method called `deinitialize` which gets run when they're disabled, but that got replaced by an `uninitialize` method using a more standard name. This doesn't actually name the method, but it's probably good to not potentially lead people to using the method.

#### Related Pull Requests
https://github.com/mattermost/mattermost/pull/29386


cwarnermm (approved):
Thanks, @hmhealey!!

amyblais (approved):

PR #1452, opened 2025-05-20
Add parallel tests guidelines

streamer45:
#### Summary

Before merging https://github.com/mattermost/mattermost/pull/30078 I'd like to ensure we have some documented guidelines on how/when to use parallelism in unit/integration tests for our server code. 

This is what I have so far. Let me know if you have suggestions or questions.

### DIFF
--- a/site/content/contribute/more-info/server/tests.md
+++ b/site/content/contribute/more-info/server/tests.md
@@ -21,10 +21,79 @@ A flaky test is one that exhibits both passing and failing results when run mult
    - Copy the test failure message into the JIRA ticket description
    - Add the `flaky-test` and `triage-global` labels
    - Create a PR to skip the test by adding:
+
      ```go
      t.Skip("https://mattermost.atlassian.net/browse/MM-XXXXX")
      ```
+
      where MM-XXXXX is your JIRA ticket number
    - Link the JIRA ticket in the skip message for tracking
 
 This process helps us track and systematically address flaky tests while preventing them from blocking development work.
+
+## Writing Parallel Tests
+
+Leveraging parallel tests can drastically reduce execution time for entire test packages, such as `api4` and `app`, which are notably heavy with hundreds of tests. However, careful implementation is essential to ensure reliability and prevent flakiness. Follow these guidelines when writing parallel tests:
+
+### Enabling Parallel Tests
+
+Ensure parallel tests run behind an explicit environment flag to control execution:
+
+```go
+// In api4 and app packages, check mainHelper.Options.RunParallel
+if mainHelper.Options.RunParallel {
+    t.Parallel()
+}
+
+// In sqlstore package, a global enableFullyParallelTests variable is used instead
+if enableFullyParallelTests {
+    t.Parallel()
+}
+```
+
+To enable parallel execution, you should set the `ENABLE_FULLY_PARALLEL_TESTS` environment variable. Example:
+
+```bash
+ENABLE_FULLY_PARALLEL_TESTS=true go test -v ./api4/...
+```
+
+### When to Use Parallel Tests
+
+- **Generally Safe**: Tests with dedicated setup functions that ensure independence from other tests.
+- **Subtests**: Only safe if each subtest features its own setup function, ensuring they are decoupled and independent of execution order.
+- **Unsafe**: When a subtest depends on state changes made by another subtest, thus coupling their execution order.
+
+### Common Issues That Break Parallel Safety
+
+#### Global State
+
+Avoid reliance on global variables and registrations such as:
+
+- `LicenseValidator`
+- `platform.RegisterMetricsInterface`
+- `platform.PurgeLinkCache`
+- `model.BuildEnterpriseReady`
+- `jobs.DefaultWatcherPollingInterval`
+
+#### Filesystem Operations
+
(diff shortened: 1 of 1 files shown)


isacikgoz (approved):
💯

site/content/contribute/more-info/server/tests.md
```diff
@@ -21,10 +21,79 @@ A flaky test is one that exhibits both passing and failing results when run mult
...
+```go
+// In api4 and app packages, check mainHelper.Options.RunParallel
+if mainHelper.Options.RunParallel {
+    t.Parallel()
+}
+
+// In sqlstore package, a global enableFullyParallelTests variable is used instead
+if enableFullyParallelTests {
+    t.Parallel()
+}
```
lieut-data, line 51: Is this something each test has to do for itself? Could we add a `Parallel` method to the commonly used `testHelper` struct that encodes this logic instead?

site/content/contribute/more-info/server/tests.md
```diff
@@ -21,10 +21,79 @@ A flaky test is one that exhibits both passing and failing results when run mult
...
+```
+
+To enable parallel execution, you should set the `ENABLE_FULLY_PARALLEL_TESTS` environment variable. Example:
+
+```bash
+ENABLE_FULLY_PARALLEL_TESTS=true go test -v ./api4/...
```
lieut-data, line 57: Can we prefix this with `MM_` to keep the namespacing sane?

site/content/contribute/more-info/server/tests.md
streamer45: I'm afraid so, as I don't think that adding it to the helper would be a safe default because whether parallelism can be used depends on the specific test implementation.

site/content/contribute/more-info/server/tests.md
```diff
@@ -21,10 +21,79 @@ A flaky test is one that exhibits both passing and failing results when run mult
...
+
+### When to Use Parallel Tests
+
+- **Generally Safe**: Tests with dedicated setup functions that ensure independence from other tests.
+- **Subtests**: Only safe if each subtest features its own setup function, ensuring they are decoupled and independent of execution order.
+- **Unsafe**: When a subtest depends on state changes made by another subtest, thus coupling their execution order.
+
+### Common Issues That Break Parallel Safety
+
+#### Global State
```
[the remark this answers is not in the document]
streamer45: We do, each parallel execution of a test will have its own database, taken from a shared pool, then cleaned up before another use.

site/content/contribute/more-info/server/tests.md
lieut-data: No, what I mean is can we expose this:
```go
func (th *testHelper) Parallel(t) {
    if mainHelper.Options.RunParallel {
        t.Parallel()
    }
}
```

and just do this from the test?
```go
func TestBlah(t *testing.T) {
    th := setupHelper()
    th.Parallel(t)
    // ...
}
```

site/content/contribute/more-info/server/tests.md
lieut-data: I'm thinking of examples like `MM_DEBUG` for plugins, but happy to also change perspective. Who is the target audience for confusion?

site/content/contribute/more-info/server/tests.md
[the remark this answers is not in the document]
lieut-data: Aha, interesting, I assume that's something new with your overall set of changes? Can you point me at the pooling logic?

site/content/contribute/more-info/server/tests.md
streamer45: This is meant as a developer setting, but mostly for CI. We do use the `CI_` prefix for a few things, but I also wanted to give the opportunity to run the tests on a local environment now that it doesn't take hours to complete them so I ended up prefix-less :p

I am totally happy to add the `MM_` prefix, just sharing some context.

site/content/contribute/more-info/server/tests.md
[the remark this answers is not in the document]
streamer45: @lieut-data Here you go -> https://github.com/mattermost/mattermost/pull/30078#issuecomment-2627852044

site/content/contribute/more-info/server/tests.md
lieut-data: I'm 0/5... if you can write down your score I'll happily defer :P

lieut-data (approved):
Nice, thanks for the extra context. I'm good however you choose to address the above feedback :)

site/content/contribute/more-info/server/tests.md
streamer45: @lieut-data For simplicity (there are hundreds of files to modify as a result of this), I propose adding the helper function to `mainHelper` as well, so we can do:

```go
func TestExample(t *testing.T) {
  mainHelper.Parallel(t)
}
```

Alternatively, we can also do what you suggested:

```go
func TestExample(t *testing.T) {
  th := Setup(t)
  th.Parallel(t)
}
```

But it just becomes a pain to update automatically now since pattern matching becomes fairly complex.

streamer45: I pushed these changes at https://github.com/mattermost/mattermost/pull/30078

streamer45:
All yours @cwarnermm

cwarnermm (approved):

PR #1453, opened 2025-05-29
Modified the 'Type=oneshot' to 'Type=simple' in the systemd service file

1kjnv:
#### Summary

Modified the `'Type=oneshot'` to `'Type=simple'` in the systemd service file

When the process is started using `'systemctl start'` or `'systemctl restart'`, the status remains stuck at 

> 'activating' 

because the oneshot type does not indicate to systemd that the process has started successfully.

Changing the type to simple ensures that systemd recognizes the process as running once it is launched.

### DIFF
--- a/site/content/contribute/more-info/mobile/push-notifications/service.md
+++ b/site/content/contribute/more-info/mobile/push-notifications/service.md
@@ -41,7 +41,7 @@ For the sake of making this guide simple we located the files at `/home/ubuntu/m
     Description=Mattermost Push Notification Service
 
     [Service]
-    Type=oneshot
+    Type=simple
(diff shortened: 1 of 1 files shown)


larkox (approved):
The reasoning looks good to me, but it is completely out of my comfort zone. I asked @enahum review just in case.

enahum:
I think we want to keep oneshot as we want to wait for the service to have actually started, the option "simple" does not wait even if there was a failure.

Here is a reference that describes the difference between them https://stackoverflow.com/a/39050387

enahum:
this has been sitting here for a long while, I'm going to close it now, feel free to re-open if still wanted

PR #1460, opened 2025-07-23
Edit on GitHub: Take API docs visitors directly to API YAML source

cwarnermm:
Used Claude to update the ``page-edit.html`` partial to use the new URL (https://github.com/mattermost/mattermost/tree/master/api/v4/source) specifically for API documentation pages, while keeping the original URL for all other pages.


esarafianou (approved):
Thanks Carrie!

PR #1516, opened 2026-07-08
Update docs to include collapsible elements

krd311:
#### Summary
Adds in documentation and instructions regarding the new collapsible elements in interactive dialogs.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-69315


sbishel (approved):
Need to update version, since it was delayed. Otherwise its good.

